PR #21, opened 2015-04-02
Adding Robolectric test framework and JaCoCo code coverage reports.

tonyr59h:
This will enable us to write unit tests and figure out what needs testing via code coverage reports.

I've written some tests as an example. Here's the generated code coverage report:

We probably don't want to spend resources testing the legacy editor so we may want to add it to the excluded list [here](https://github.com/wordpress-mobile/WordPress-Editor-Android/commit/200efb8a597e1745ea755e4e08df98e6d46e4a70#diff-ff1371cc0d469a8fc19b088d1540a921R137).


WordPressEditor/build.gradle
```diff
@@ -4,10 +4,13 @@ buildscript {
     }
     dependencies {
         classpath 'com.android.tools.build:gradle:1.0.0'
+        classpath 'org.robolectric:robolectric-gradle-plugin:0.14.1'
```
fadils, line 7: @tonyr59h this gradle plugin is deprecated. You might want to update `build:gradle` to at least `1.1` so that Robolectric can be integrated seamlessly.

WordPressEditor/build.gradle
```diff
@@ -41,6 +44,17 @@ dependencies {
     compile 'org.wordpress:analytics:1.0.0'
     releaseCompile project(path:':libs:utils:WordPressUtils', configuration: 'release')
     debugCompile project(path:':libs:utils:WordPressUtils', configuration: 'debug')
+
+    // Test libraries
+    testCompile 'junit:junit:4.10'
+    testCompile 'org.mockito:mockito-core:1.10.19'
+    testCompile 'org.robolectric:robolectric:2.4'
```
fadils, line 50: This version is not compatible with `appCompat` but probably irrelevant for now. Any possibility where `appCompat` comes into play in the project?

aforcier:
Looks good @tonyr59h! We just merged #22, which fixes Travis integration among other things - could you merge develop in when you have a chance?

@fadils: most of what I could find re: `Robolectric 2.4` and `AppCompat` problems seem related to testing `ActionBarActivity` directly, which shouldn't affect testing the new editor as far as I can tell just now. Definitely something to keep an eye on though.

fadils:
:shipit:

tonyr59h:
Looks like updating the build tools breaks tests. And I don't think Travis likes the [workaround](https://github.com/wordpress-mobile/WordPress-Editor-Android/pull/21/files#diff-ff1371cc0d469a8fc19b088d1540a921R56) added for better IDE support. Can we disable that lint check?

fadils:
@tonyr59h for `build.gradle:1.1+` robolectric-gradle-plugin is no longer needed..
Also, can we set `abortOnError` as `false`?

tonyr59h:
> @tonyr59h for build.gradle:1.1+ robolectric-gradle-plugin is no longer needed..

I tried getting the latest version to recognize and run the tests but didn't have any luck, I'm happy to leave it for another PR.

aforcier:
@tonyr59h I disabled the `InvalidPackage` lint check for Robolectric, looks like all is well without it.

I also tried updating the build tools and couldn't get the tests to run either. Definitely something we can work in at a later time. For now tests work and Travis is happy.

:shipit:

Issue #131, opened 2015-04-20
Animate format bar selection


aforcier:
Example: https://cloudup.com/cb2NTJ_8DxQ

thianhlu:
related to #145

bummytime:
iOS Animation implementation was here: https://github.com/wordpress-mobile/WordPress-Editor-iOS/pull/678

aforcier:
First pass:

https://cloudup.com/cpuLSkcC5py

What do you think @bummytime @thianhlu?

thianhlu:
:+1:  you can also remove underline at this point as well

Issue #127, opened 2015-04-13
Migrate updated blockquote logic from iOS


bummytime:
After [this PR](https://github.com/wordpress-mobile/WordPress-Editor-iOS/pull/622) lands, update the blockquote logic in the Android editor.

aforcier:
Fixed in #153.

Issue #123, opened 2015-04-08
Unable to apply style mid-typing (autocorrect, API19+)


aforcier:
Currently, applying a style (e.g. bold) while typing a word (i.e., not right after a space or line break) doesn't work on software keyboards on API 19+. This seems to be an issue with how Android's autocorrect behaves inside an API 19+ Chromium `WebView`.

Notes:
- This is not an issue on API 18 and lower. In fact, API 18 (and probably lower also) displays suggestions correctly even as formatting is applied
- A possible workaround involves selectively disabling autocorrect when applying styling mid-word. However, this approach only works for `API>17`, and doesn't work for IME keyboards or the Swype keyboard (most likely others as well).

aforcier:
The `test-bold-input` branch has a partial demo of the workaround.

bummytime:
Tested: also fails on Samsung S6 (API 21)

thianhlu:
Other downsides are trying to apply additional styles mid-word or removing styles mid-word.

rachelmcr:
This was found again in alpha testing by Mo, Moto X 2, Android 6.0:

> Related, once you switch something to bold (Selct word and tap B), toggling it off and typing reverts back to bold unless you type a space first and then start typing again.
> 1. type a word
> 2. select it
> 3. tap B to bold it
> 4. tap B again to disable bold
> 5. type another letter
> 
> That letter will be bold but probably shouldn’t.

It's also a problem if you try to remove styles at the end of a word, but before typing a space. As tested on my Nexus 9, Android 6.0.1:
1. Tap B
2. Type a word
3. Tap B (at the end of the word, before the space) -> the B button is no longer highlighted
4. Type a space -> the B button is highlighted again
5. Continue typing and the next word is bold

rachelmcr:
This came up again with a user trying out the latest 5.4 release. Her steps:
1. Tap the italics format button
2. Type a word
3. Tap the italics format button again
4. Type a space and another word

She was surprised that the second word was still in italics. She expected to be able to disable italics immediately after the end of the first word. Her reasoning for disabling italics before typing a space: If she types a space and then disables italics, then if she goes back to add text there later it will end up italicized when she doesn't want it to be.

She also noted that there wasn't enough color contrast between the enabled and disabled format buttons for her to see that the italics formatting was still enabled.

aforcier:
[...]

Issue #10, opened 2015-03-30
Format bar (phone)


bummytime:
On the iOS editor, the iPhone format bar buttons were updated to the following:

![screen shot 2015-03-30 at 11 09 00 am](https://cloud.githubusercontent.com/assets/154014/6900631/7478c69e-d6cd-11e4-9ff9-27b8967b9444.png)

This was a completely data-driven decision based on usage metrics. Not sure if Android users will have different needs, but we should find out. Also note how the layouts differ between tablet and phone.

For context, the current editor has the following:

![screen shot 2015-03-30 at 10 57 33 am](https://cloud.githubusercontent.com/assets/154014/6900664/9c1f86ce-d6cd-11e4-9849-f1fa121ce227.png)

/cc @thinahlu @meremagee

aforcier:
Some notes:
- As on iOS, the HTML button should always be visible, while the rest of the buttons scroll (if needed).
- If not scrolling, buttons should be evenly spaced.
- On iOS, the format bar icons gray out (except for the HTML button) when editing the title

bummytime:
Another note, if a hardware keyboard is being used, format bar should be placed at the bottom of the screen (assuming the soft keyboard is not displayed).

thianhlu:
@thianhlu 

Do review on icons and match material design enabled/disabled states

thianhlu:
lets leave buttons and disabled button state the same as iOS.

We could however add a slight animation for tapping buttons, https://cloudup.com/cb2NTJ_8DxQ, which could help provide a bit more visual feedback but could be overkill.

aforcier:
Most of the buttons aren't hooked up to anything yet, but here's a first look at the re-styled format bar on Android (phone-sized):

![2015-04-17-format-bar-preview-cropped-full](https://cloud.githubusercontent.com/assets/9613966/7204747/a368609e-e4f2-11e4-9063-f2c57676522b.png)

cc @thianhlu

aforcier:
And here it is with the less flattering Jelly Bean default keyboard:

![api17-format-bar](https://cloud.githubusercontent.com/assets/9613966/7205407/20a86762-e4f7-11e4-9608-6e57ccbcd60d.png)

cc @thianhlu

thianhlu:
Dumb question @aforcier. What's the OS usage for our android devices? Are most people on Lollipop or Kitkat at this point? How does it look on those default keyboards?

meremagee:
[...]

PR #133, opened 2015-04-21
Issue #10: Format bar (phone)

aforcier:
Addresses #10.
- Updated the format bar appearance to use the iOS assets and styling
- Implemented `bold`, `italic`, `blockquote`, `orderedList`, and `unorderedList` formatting
- Added `media`, `link` and `HTML` buttons (currently inactive), to be implemented in future PRs
- Format bar buttons (except `HTML`) scroll horizontally when needed
- Focusing on the title disables the format bar (except `HTML`)

2015-04-17-format-bar-preview-cropped-full

Some notes:
- There are a few bugs present:
  1. Android's autocorrect/autocomplete in some cases interferes when applying/removing formatting (`blockquote`/`ol`/`ul`) and causes bugginess (duplicating words mainly). This is an extension of issue #123.
  2. Blockquote behaves a bit strangely, with the same issues listed in https://github.com/wordpress-mobile/WordPress-Editor-iOS/issues/597, which will be fixed in #127.
- Possible animation for format bar buttons (and removal of the blue underline) sold separately (#131)


aforcier:
Note that on small screens a format bar button can be partially hidden (in this example the orderedList icon is barely making into the visible part of the scrollable format bar):

format-bar-scroll

Should we treat this case differently? cc @thianhlu

aforcier:
cc @bummytime

bummytime:
@aforcier re: the clipped format bar above, if we tightened up the static HTML button on the right (make its width smaller), we would see more of that obscured link icon. 

That "HTML" text is from the iPad version...on the phone it looks a little different: https://cldup.com/p9IQoNoLwT.jpg

aforcier:
@bummytime I don't see an asset for that here:

https://github.com/wordpress-mobile/WordPress-Editor-iOS/tree/develop/Assets

Is it a text button on iPhone then?

bummytime:
Yes. :-)

bummytime:
Though we don't have to do the same thing on Android...

aforcier:
Here's a pass at tightening up the HTML button:

format-bar-scroll-small-html

Comparison with earlier version:

format-bar-scroll-comparison

What do you think? cc @thianhlu

thianhlu:
I like.

aforcier:
Updated with shrunken HTML button. Slightly different from the above sample screenshots (I used text in those, while my last change and the screenshots below use an image resource).

Here's what the format bar looks like now on a Nexus 4 (left) and a Nexus One running ICS (right):

format-bar-scroll-small-html-nexusone-nexus4-comparison

(full Nexus4 screenshot: https://cloudup.com/cYEjCK0zfJ9)

bummytime:
Looking good @aforcier !

Could we expand the scroll area to the left edge of the screen?

napkin 3 04-24-15 10 43 53 am

bummytime:
@thianhlu More of a design question here. Looking at @aforcier's image cap from above:

screen shot 2015-04-24 at 10 48 26 am

It seems like we need some kind of graphical affordance to let the user know that the toolbar can scroll. Evernote achieves this by adding a material-like drop shadow to the static part of the toolbar on the right hand side:

screen

Not saying we have to do that, but I think we should do something...

bummytime:
@aforcier ok. I chatted with @thianhlu offline and I will make a separate issue for the scrolling UX. So in terms of this PR, if you fix the left hand scroll padding I mentioned above, I would say :shipit:!

aforcier:
Fixed left edge scroll issue:

format-bar-left-scroll-fix

bummytime:
:shipit:

PR #138, opened 2015-04-28
Separate inline JS/CSS from the main editor html file

### LINKED ISSUE #124: Separate CSS, JS, and HTML into separate files
Much like this [PR](https://github.com/wordpress-mobile/WordPress-Editor-iOS/pull/618) did on the iOS side, let's remove inline JS and put all everything into separate files.

bummytime:
This PR: 
- Fixes #124 
- Upgrades JQuery to 2.1.3
- removes some unused HTML & JS files

/cc @aforcier


aforcier:
A few HTML style nits for `android-editor.html` (following [codeguide.co](http://codeguide.co)):
- drop trailing slash from meta and hr tags
- drop type specification from stylesheet declaration

We also have a spacing discrepancy in the same file, these lines use (mostly) tabs while the rest of the file uses double spaces (double spaces seem to be the standard):

https://github.com/wordpress-mobile/WordPress-Editor-Android/blob/issue/124-separate-files/libs/editor-common/assets/android-editor.html#L18-L26

Other than these nits, everything looks good, and testing on APIs 14, 17, 19, 21 revealed no new bugs (no existing issues magically fixed either). Once they're taken care of I say :shipit:!

bummytime:
@aforcier good catch with the style nits. I will update.

PR #141, opened 2015-05-01
Issue #139: WebView padding for tablets

### LINKED ISSUE #139: Add padding to WebView on tablets
Similar to iPad: https://github.com/wordpress-mobile/WordPress-Editor-iOS/pull/487

Many Android tablets are narrower than iPads, a good approach would probably be to add margins on all tablets during landscape orientation, and also on portrait orientation for large tablets only (720dp+).

aforcier:
Fixes #139.

Added side padding for better readability on tablets.

Minimum screen width to show padding is `720dp`, with more padding added at `1280dp`.

This translates to no padding on small tablets in portait mode, some padding on small tablets in landscape and large tablets in portrait, and extra padding on large tablets in landscape. Here are some samples:
### Nvidia Shield:

tablet-padding-shield-portrait

tablet-padding-shield-landscape
### Nexus 9:

tablet-padding-nexus9-portrait

tablet-padding-nexus9-landscape
### Nexus 10:

tablet-padding-nexus10-portrait

tablet-padding-nexus10-landscape


aforcier:
I used an extra css file for the Android tablet changes so they don't cause trouble if ported back to iOS. We could also similarly extract iOS-specific styles on the iOS side to keep things tidy:

https://github.com/wordpress-mobile/WordPress-Editor-Android/blob/develop/libs/editor-common/assets/editor.css#L290-L302

aforcier:
cc @bummytime, @thianhlu

thianhlu:
@aforcier Can you add a bit more space (e.g. 10px) between the buttons. Overall this looks good. I am still working on how we can define a better style for the toolbar. It just doesn't look right.

aforcier:
Added some extra margins between buttons and between groups of buttons for wide screens.

This targets screen widths of at least 800dp. I didn't use 720dp (the cut-off point for showing side margins on tablets) because adding ~10px between buttons won't fit much below 800dp. We could drop some of the extra space and close that gap if we want. Or I could refine the padding rules into more ranges.

The Nexus 9 in portrait below is a good example, at 768dp width it's right in the range that could be covered if we use a bit less spacing. What do you think @thianhlu, should it have more spacing between buttons?

Here's the ref I'm using for tablet sizes: http://developer.android.com/design/style/metrics-grids.html

### Nvidia Shield:

tablet-padding-shield-portrait-v2
(unchanged)

tablet-padding-shield-landscape-v2

### Nexus 9:

tablet-padding-nexus9-portrait-v2
(unchanged)

tablet-padding-nexus9-landscape-v2

### Nexus 10:

tablet-padding-nexus10-portrait-v2

tablet-padding-nexus10-landscape-v2

bummytime:
:shipit:

Issue #128, opened 2015-04-14
Add integration tests


aforcier:
Since `Robolectric` doesn't load a `WebKit`, it can't be used to test interactions with the JavaScript `ZSSEditor`. We'll need to add some (much slower) emulator-reliant integration tests to cover everything `Robolectric` unit tests can't cover.

PR #143, opened 2015-05-08
Issue #128: Integration tests

aforcier:
Addresses #128, adding support for integration tests using the [Android testing framework](http://developer.android.com/tools/testing/testing_android.html).

I also added a few actual tests, which test some of the `EditorFragment`'s functionality with a real, live `ZSSEditor`. There's also a class for testing the `ZSSEditor` inside an `EditorWebView`, with no UI. The only test in this class for now makes sure that startup callbacks are being sent from JavaScript.

These tests aren't part of Travis' build script yet - they have to be run manually using `gradlew connectedAndroidTest`. In my opinion, we can consider adding these to Travis when the editor is further along and we have some idea of the extent and duration of the tests.

### DIFF
--- a/WordPressEditor/src/androidTest/java/org/wordpress/android/editor/integration/EditorFragmentTest.java
+++ b/WordPressEditor/src/androidTest/java/org/wordpress/android/editor/integration/EditorFragmentTest.java
@@ -1,0 +1,64 @@
+package org.wordpress.android.editor.integration;
+
+import android.app.Activity;
+import android.test.ActivityInstrumentationTestCase2;
+import android.widget.ToggleButton;
+
+import org.wordpress.android.editor.R;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.wordpress.android.editor.TestingUtils.waitFor;
+
+public class EditorFragmentTest extends ActivityInstrumentationTestCase2<MockEditorActivity> {
+    Activity mActivity;
+    EditorFragmentForTests mFragment;
+
+    public EditorFragmentTest() {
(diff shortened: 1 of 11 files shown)


aforcier:
cc @bummytime

WordPressEditor/src/androidTest/java/org/wordpress/android/editor/integration/EditorFragmentTest.java
```diff
@@ -0,0 +1,108 @@
...
+        mFragment = (EditorFragmentForTests) mActivity.getFragmentManager().findFragmentByTag("editorFragment");
+    }
+
+    public void testDomLoadedCallbackReceived() {
+        // initJsEditor() should have been called on setup
+        assertTrue(mFragment.mInitCalled);
+
+        long start = System.currentTimeMillis();
+        while(!mFragment.mDomLoaded) {
+            waitFor(10);
```
bummytime, line 36: I see we have a few `sleep()` calls in these tests. I get why they are there, however they tend to make tests brittle (esp when running within a CI environment). Not sure if we can do anything about them now...but may be an opportunity for future refactoring.

bummytime:
Ok @aforcier, looks good. :shipit:

WordPressEditor/src/androidTest/java/org/wordpress/android/editor/integration/EditorFragmentTest.java
aforcier, line 36: Agreed, definitely something to think about.

PR #146, opened 2015-05-19
Issue/fix travis build

maxme:
Note: I opened this PR against `issue/130-landscape-viewport` because it wasn't merged in `develop` yet.


aforcier:
Looks like the 1.1.0 robolectric gradle plugin requires the test classes to be in `src/test/java` rather than `src/androidTest/java`. With that change everything works locally and the unit tests run.

With that change, :shipit:.

maxme:
renamed androidTest directory in a2a64f1
also removed `DrawerLayoutImpl.file` that was not used anymore

PR #147, opened 2015-05-19
Issue #130: Small viewport in landscape mode

### LINKED ISSUE #130: Small viewport on landscape devices
With the addition of the format bar, some phones have a very small viewport with the keyboard up:

landscape-nexus4-narrow-viewport

(Nexus4, landscape mode, standard keyboard, API 22)

One idea might be to use fullscreen mode while editing if the viewport is below a certain size. This would be tied to scrolling: fullscreen mode while scrolling down and editing, exit fullscreen and restore the action bar when scrolling up. See the Pocket app for an example.

aforcier:
Fixes #130.

The action bar is hidden while the keyboard is up and revealed again when the keyboard is dismissed. This applies on landscape mode only, and only for devices smaller than a large tablet. A large tablet in this case is any device with a width of at least `1280dp` in landscape mode  ([ref](http://developer.android.com/design/style/metrics-grids.html)).

This PR also includes some updates to the test setup, fixing an issue this PR introduced which was causing Travis CI to fail. The issue was due to the `robolectric` plugin, and was connected to `v4` support library classes being added to the integration tests. Updating the robolectric plugin version fixed the problem. Unit tests are now placed under `/src/test/java/` (to comply with the new plugin's setup requirements) and integration tests under `/src/androidTest/java/`. Thanks @maxme for the detective work!


bummytime:
Everything looks good @aforcier. I did notice the following nit on a Nexus 4 API 17 device:

1) Hold the device in portrait mode, soft keyboard down
2) Rotate the device into landscape
3) Soft keyboard is down and actionbar is hidden
4) Tap the text field
5) Then dismiss the soft keyboard
6) Action bar is now visible

To me this is not critical, but wanted to note it here. :shipit:

PR #153, opened 2015-05-21
Blockquote fixes

bummytime:


bummytime:
Stuff to fix:
- [x] Tapping blockquote on a already blockquoted line, does not do anything
- [x] Blockquote'ing a single line of text with in an enclosed tag causes bquote to be placed on new line (API  > 19)
- [x] Blockquote'ing a list is broken

bummytime:
@aforcier Can you start giving this a test run and let me know about any bugs you encounter?

aforcier:
Ran into some weirdness on an API22 Nexus 4:

https://cloudup.com/cN0k8O4nIXV

This seems to be another autocorrect-related issue - disabling autocorrect got rid of this bug. Haven't had any other issues yet (everything seems to work great with autocorrect off anyway). I'll test this on an API18 Galaxy S3 next week.

bummytime:
Nice catch @aforcier....I have not tried using autocorrect with this yet :grimacing: So in the video, are you pressing `testing` in the autocorrect bar and then the blockquote CSS goes wacky?

aforcier:
@bummytime Actually all I'm doing is hitting the blockquote button after I've finished typing `Testing` - I didn't have to select `Testing` from the suggestions bar. The wacky switch from correct blockquote display to it suddenly shrinking happens without any user input.

The duplication of `Testing` at the end happens after pressing space, which happens for other HTML formatting as well (so it's not specific to blockquotes).

aforcier:
Also seeing some weirdness on the Samsung S3 (API 18) when using lists inside blockquotes. Here's a demo:

https://cloudup.com/coESBG3V_tj

aforcier:
Here's a simpler case of weirdness on the Samsung S3 (not involving nested lists). It happens when you apply formatting (bold, italics) to text inside a list within blockquotes:

https://cloudup.com/cvtjD4Blk9Y

Basically it looks like when you press the list button again, the list is taken away but the text is indented.

bummytime:
> Ran into some weirdness on an API22 Nexus 4...

Looks like the blockquote node is being turned into a span :-/

screen shot 2015-06-09 at 8 40 11 pm

bummytime:
@aforcier I am pulling the autocorrect issue out of this PR here: #159

aforcier:
The only remaining issue with this PR (with the autocorrect issue having been extracted to #159) is the glitchy behavior of lists within blockquotes on the Samsung S3 (`API 18`). Since this doesn't happen on the Samsung S6 (`API 21`), let's open a separate issue for it and track it as more Samsung devices run the editor leading up to release.

Other than that, tested on `API 14+` emulators and devices, and everything looks great.

:shipit:

aforcier:
Note that this PR also seems to fix issue #160 - the methods causing the issue have been refactored.

PR #154, opened 2015-06-03
Caret should now be visible when soft keyboard displays

### LINKED ISSUE #122: Auto-scroll webview when bringing up keyboard
Title has it - the editor should scroll so that the keyboard doesn't hide the line being edited.

bummytime:
Title has it. Fixes #122

/cc @aforcier


aforcier:
Nice fix @bummytime! Working great on API 16+ emulators, API 22 devices and an API 18 Galaxy S3.

However, on API 14-15 the WebView doesn't scroll at all when the soft keyboard displays. I think this has always been the case and it's probably a completely different issue. Up to you if you'd prefer to address that here or open a separate issue for API 14-15 scrolling.

bummytime:
Thanks @aforcier. Nice catch on the API 14-15 thing...I thought I tested this on there, but obviously did not. :-)  I think I would like to merge this in if you are ok with it and then create a separate issue for the scrolling on those API levels.

aforcier:
Sounds good to me!

:shipit:

Issue #177, opened 2015-07-07
Style text in HTML mode


aforcier:
Add styling to HTML tags/attributes/entities for HTML mode (ref #11).

iOS HTML mode:

![iOS editor HTML mode 1](https://cloud.githubusercontent.com/assets/154014/6904990/51567eb4-d6ea-11e4-8c6d-b4bf8df7cc8d.png)

![iOS editor HTML mode 2](https://cloud.githubusercontent.com/assets/154014/6905010/8a1cb6b4-d6ea-11e4-8a74-5303c4b6e842.png)

This also includes automatically adding and removing styling as the user modifies the content.

Issue #11, opened 2015-03-30
HTML mode


bummytime:
What pairs nicely with a new 100% visual editor? Why, an HTML mode of course. On iOS we have this:

![scaled screen shot 2015-03-30 at 2 36 55 pm](https://cloud.githubusercontent.com/assets/154014/6904990/51567eb4-d6ea-11e4-8c6d-b4bf8df7cc8d.png)

![scaled screen shot 2015-03-30 at 2 38 53 pm](https://cloud.githubusercontent.com/assets/154014/6905010/8a1cb6b4-d6ea-11e4-8a74-5303c4b6e842.png)

Things to consider:
- Paragraph formatting - show spaces instead of `<p>`'s. (Should be handed [here](https://github.com/wordpress-mobile/WordPress-Editor-iOS/blob/3fb089aec656f99e35447ca21ace89eea254181a/Assets/wpsave.js) and [here](https://github.com/wordpress-mobile/WordPress-Editor-iOS/blob/3fb089aec656f99e35447ca21ace89eea254181a/Assets/wpload.js).)
- Fontface (Menlo-Regular 14pt on iOS)
- Tag and element colors
- Post title positioned in the same place as when in visual mode (with the same font...Merriweather)
- Format bar button should act as a toggle between modes

@thinahlu can offer up more specifics if he wants.

Ref: https://github.com/wordpress-mobile/WordPress-Editor-iOS/issues/110
and https://github.com/wordpress-mobile/WordPress-Editor-iOS/issues/154

aforcier:
I ran into a difference in how the current iOS editor and the legacy Android editor handle formatting in HTML mode that I'd like some input on.

Here is how applying formatting with no text selected works on each.

iOS: https://cloudup.com/c_D1PfneupY
- Press format button
- Opening and closing tags are inserted simultaneously
- Format button is immediately un-highlighted
- Cursor is placed at the end of the closing tag, and you have to move the cursor back to between the tags to type your text that should be formatted

Android (legacy): https://cloudup.com/coR6KC0HIzz
- Press format button
- Opening tag only is inserted
- Format button is now highlighted
- Cursor is placed at the end of the opening tag, ready for text to be inserted and formatted right away
- When finished, press format bar button again
- Closing tag is inserted, and the button is un-highlighted

My vote would be to keep the legacy Android method for the new editor. What do you think @bummytime @meremagee @thianhlu?

bummytime:
[...]

PR #184, opened 2015-07-15
Issue #177: HTML mode styling

aforcier:
Addresses #177, styling HTML elements when in HTML mode. Uses the same style rules as [iOS](https://github.com/wordpress-mobile/WordPress-Editor-Android/issues/11).

177-styling-screens

### DIFF
--- a/WordPressEditor/src/main/java/org/wordpress/android/editor/HtmlStyleTextWatcher.java
+++ b/WordPressEditor/src/main/java/org/wordpress/android/editor/HtmlStyleTextWatcher.java
@@ -78,7 +78,11 @@ public class HtmlStyleTextWatcher implements TextWatcher {
         int closingTagLoc = mModifiedText.toString().indexOf(closingSymbol);
         int openingTagLoc = content.toString().lastIndexOf(openingSymbol, mStart + closingTagLoc);
         if (openingTagLoc > 0) {
-            updateSpans(content, openingTagLoc, mStart + closingTagLoc + 1);
+            if (mStart + closingTagLoc < content.length()) {
+                updateSpans(content, openingTagLoc, mStart + closingTagLoc + 1);
+            } else {
+                updateSpans(content, openingTagLoc, content.length());
+            }
         }
     }
 
--- a/WordPressEditor/src/main/java/org/wordpress/android/editor/HtmlStyleTextWatcher.java
+++ b/WordPressEditor/src/main/java/org/wordpress/android/editor/HtmlStyleTextWatcher.java
@@ -42,27 +42,34 @@ public class HtmlStyleTextWatcher implements TextWatcher {
             return;
         }
 
+        SpanRange spanRange;
+
         // If the modified text included a tag or entity symbol ("<", ">", "&" or ";"), find its match and restyle
         if (mModifiedText.toString().contains("<")) {
-            restyleForChangedOpeningSymbol(s, "<");
+            spanRange = restyleForChangedOpeningSymbol(s, "<");
         } else if (mModifiedText.toString().contains(">")) {
-            restyleForChangedClosingSymbol(s, ">");
+            spanRange = restyleForChangedClosingSymbol(s, ">");
         } else if (mModifiedText.toString().contains("&")) {
-            restyleForChangedOpeningSymbol(s, "&");
+            spanRange = restyleForChangedOpeningSymbol(s, "&");
         } else if (mModifiedText.toString().contains(";")) {
-            restyleForChangedClosingSymbol(s, ";");
+            spanRange = restyleForChangedClosingSymbol(s, ";");
         } else {
             // If the modified text didn't include any tag or entity symbols, restyle if the modified text is inside
             // a tag or entity
-            if (!restyleNormalTextIfWithinSymbols(s, "<", ">")) {
-                restyleNormalTextIfWithinSymbols(s, "&", ";");
+            spanRange = restyleNormalTextIfWithinSymbols(s, "<", ">");
+            if (spanRange == null) {
+                spanRange = restyleNormalTextIfWithinSymbols(s, "&", ";");
             }
+        }
+
+        if (spanRange != null) {
+            updateSpans(s, spanRange);
         }
 
         mModifiedText = null;
     }
 
-    private void restyleForChangedOpeningSymbol(Editable content, String openingSymbol) {
+    protected SpanRange restyleForChangedOpeningSymbol(Editable content, String openingSymbol) {
         String closingSymbol = getMatchingSymbol(openingSymbol);
 
         // Apply span from the first added/deleted opening symbol until the closing symbol in the content matching the
@@ -73,11 +80,12 @@ public class HtmlStyleTextWatcher implements TextWatcher {
 
         int closingTagLoc = content.toString().indexOf(closingSymbol, lastOpeningTagLoc);
         if (closingTagLoc > 0) {
-            updateSpans(content, firstOpeningTagLoc, closingTagLoc + 1);
+            return new SpanRange(firstOpeningTagLoc, closingTagLoc + 1);
         }
+        return null;
     }
 
-    private void restyleForChangedClosingSymbol(Editable content, String closingSymbol) {
+    protected SpanRange restyleForChangedClosingSymbol(Editable content, String closingSymbol) {
         String openingSymbol = getMatchingSymbol(closingSymbol);
 
         int firstClosingTagInModLoc = mOffset + mModifiedText.toString().indexOf(closingSymbol);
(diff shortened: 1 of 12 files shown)


aforcier:
cc @bummytime, @thianhlu

bummytime:
@aforcier Could we turn off spell check in the HTML EditView? Kind of distracting seeing all those red lines :-) (We have it off in iOS)

screen shot 2015-07-15 at 11 10 40 am

e.g. `android:inputType="textNoSuggestions"`

thianhlu:
looks good @aforcier. I'll take a further look once merged and deployed to hockey

bummytime:
@aforcier looks like we are not persisting post updates between orientation changes. We can make a separate issue for this outside of this PR since it affects the visual editor too.

bummytime:
Tested on:
- API 16 :+1: 
- API 17 :+1: 
- API 18 :+1: 
- API 19 :+1: 
- API 21 :-1:  — More below 
- API 22 :+1:  (including Samsung S6)

For the API 21 issues, check out this video: https://cloudup.com/cq0LyzaT5Os

I did not dig into what is precisely happening, just saw the strange behavior setting the title and switching modes. Also, FWIW, in API 21, it seemed like there was a lot of latency when typing. None of the other APIs exhibited these 2 behaviors.

WordPressEditor/src/main/java/org/wordpress/android/editor/HtmlStyleTextWatcher.java
```diff
@@ -0,0 +1,196 @@
...
+                    mLastOperation = Operation.INSERT;
+                    mOffset = start;
+                    mModifiedText = s.subSequence(start + before, start + count);
+                }
+            }
+        }
+    }
+
+    @Override
+    public void afterTextChanged(Editable s) {
```
bummytime, line 66: Do we need to make sure `s` is not null or empty and, if so, just `return`?

WordPressEditor/src/main/java/org/wordpress/android/editor/HtmlStyleTextWatcher.java
```diff
@@ -0,0 +1,196 @@
...
+
+        if (spanRange != null) {
+            updateSpans(s, spanRange);
+        }
+
+        mModifiedText = null;
+        mLastOperation = Operation.NONE;
+    }
+
+    protected SpanRange restyleForChangedOpeningSymbol(Editable content, String openingSymbol) {
```
bummytime: I don't want to be "that guy", but could we add some javadoc comments to all methods that are not `@Override`? Technically in our (Android's) style guidelines.

bummytime: I won't leave any other comments about javadoc'ing methods so I don't clutter up this PR. :-)

WordPressEditor/src/main/java/org/wordpress/android/editor/HtmlStyleUtils.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.regex.Pattern;
+
+public class HtmlStyleUtils {
+    public static final int TAG_COLOR = Color.rgb(0, 80, 130);
+    public static final int ATTRIBUTE_COLOR = Color.rgb(158, 158, 158);
+
+    public static final String REGEX_HTML_TAGS = "(<\\/?[a-z][^<>]*>)";
+    public static final String REGEX_HTML_ATTRIBUTES = "(?<==)('|\")(.*?\\1)(?=.*?>)";
+    public static final String REGEX_HTML_COMMENTS = "(<!--.*?-->)";
+    public static final String REGEX_HTML_ENTITIES = "(&#34;|&#38;|&#39;|&#60;|&#62;|&#160;|&#161;|&#162;|&#163;" +
```
bummytime, line 24: WOW! Big list of entities. :dancers:

WordPressEditor/src/main/java/org/wordpress/android/editor/HtmlStyleTextWatcher.java
```diff
@@ -0,0 +1,196 @@
...
+                mOffset = start;
+                mModifiedText = s.subSequence(start + after, start + count);
+            }
+        }
+    }
+
+    @Override
+    public void onTextChanged(CharSequence s, int start, int before, int count) {
+        int lastCharacterLocation = start + count - 1;
+        if (s.length() > lastCharacterLocation) {
```
bummytime, line 49: If `s` is null, we would get a NPE. Any chance of that here on this `@Override`

WordPressEditor/src/main/java/org/wordpress/android/editor/HtmlStyleTextWatcher.java
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    private int mOffset;
+    private CharSequence mModifiedText;
+    private Operation mLastOperation;
+
+    @Override
+    public void beforeTextChanged(CharSequence s, int start, int count, int after) {
+        int lastCharacterLocation = start + count - 1;
+        if (s.length() > lastCharacterLocation && lastCharacterLocation >= 0) {
```
bummytime, line 26: If `s` is null, we would get a NPE. Any chance of that here on this `@Override`

WordPressEditor/src/main/java/org/wordpress/android/editor/HtmlStyleUtils.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    public static void styleHtmlForDisplay(Spannable content, int start, int end) {
+        applyRegex(content, start, end, REGEX_HTML_TAGS);
+        applyRegex(content, start, end, REGEX_HTML_ATTRIBUTES);
+        applyRegex(content, start, end, REGEX_HTML_COMMENTS);
+        applyRegex(content, start, end, REGEX_HTML_ENTITIES);
+    }
+
+    private static void applyRegex(Spannable content, int start, int end, String regex) {
```
bummytime: Might be a good idea to do some input param validation here. the pattern matcher will blow chunks if a null value is passed in.

bummytime:
Ok @aforcier, my first pass through is done — very nice job overall.  :cake:  I :heart: all of the tests you put in here (all of which are green for me).

WordPressEditor/src/main/java/org/wordpress/android/editor/HtmlStyleUtils.java
aforcier, line 24: Taken verbatim from [iOS](https://github.com/wordpress-mobile/WordPress-Editor-iOS/blob/a983547b5724d5fc3b79865da83e81db2b0b9de4/Classes/ZSSTextView.m#L59-L60) :wink:

aforcier:
Thanks for the thorough feedback @bummytime!

> Could we turn off spell check in the HTML EditView? Kind of distracting seeing all those red lines :-) (We have it off in iOS)

Definitely agree with that, will add it in.

> looks like we are not persisting post updates between orientation changes. We can make a separate issue for this outside of this PR since it affects the visual editor too.

As we discussed elsewhere, the example app in this repo is setup to reload the editor from scratch on orientation change. The editor itself supports either situation. Inside WPAndroid the activity is setup to persist changes on rotation, and that's actually working for HTML mode as well as visual from my testing.

> API 21 issues

Looking into that, could be an emulator issue.

Checklist:
- [x] disable spellcheck for HTML mode content
- [x] look into API21 issue
- [x] add some javadoc comments
- [x] add some extra null checks/input validation as suggested

Note: it's unclear whether the `CharSequence`/`Editable` passed to `beforeTextChanged()`, `onTextChanged()`, and `afterTextChanged()` can ever be null, adding in null checks just in case.

bummytime:
Rock on @aforcier!

aforcier:
Update: Made all the changes in the checklist above.

Note about the API 21 bug:

I was able to reproduce the issue (using both Geny and the stock emulator). It seems to be a threading bug internal to `API 21` and Chromium: a race condition occurs somewhere in the UI thread <-> Chromium interaction, and this causes `getTitle()`/`getContent()` to never update the title/content and pass it to HTML mode.

The good news is, it looks like **this has been fixed** in Android 5.1 **as well as in Chromium 40** (current version `43`), which Android 5.0 should auto-update to (otherwise it can be done by visiting the app store):

https://code.google.com/p/android/issues/detail?id=79924#c28

That looks like our bug, unfortunately the API 21/22 emulators run Chromium `37` and it can't be updated, so this would need to be confirmed on a physical device.

aforcier:
Ready for another round of feedback @bummytime :tada:

bummytime:
> That looks like our bug, unfortunately the API 21/22 emulators run Chromium 37 and it can't be updated, so this would need to be confirmed on a physical device.

Nice sleuthing @aforcier!

bummytime:
Your changes are great @aforcier...thank you!

:tada: :dancer: :nut_and_bolt: :ear: :earth_americas: :shipit:

Issue #179, opened 2015-07-09
WebView loses focus switching from HTML > visual with HW keyboard on API<19


aforcier:
On `API 14-18`, when using a hardware keyboard, switching from HTML to visual mode causes the `WebView` to lose focus, and you have to tap to be able to continue typing.

When this happens this JS error appears in the log:

`Uncaught Error: INDEX_SIZE_ERR: DOM Exception 1 -- From line 1953 of file:///android_asset/ZSSRichTextEditor.js` [(link)](https://github.com/wordpress-mobile/WordPress-Editor-Android/blob/af99a4faa0b46ee87a74530f8c02ef1024e62bdf/libs/editor-common/assets/ZSSRichTextEditor.js#L1953)

Looks like the JS side is having trouble handling the `WebView` being unfocused.

`API<19` suggests that some eccentricity of the pre-Chromium browser might be part of the problem.

(ref #176)

aforcier:
Fixed in https://github.com/wordpress-mobile/WordPress-Editor-Android/pull/326.

PR #189, opened 2015-07-17
Issue #171: Show placeholder text

### LINKED ISSUE #171: Placeholder text not displaying
Nexus 9

aforcier:
Fixes #171. Displays placeholder text for empty fields.

171-placeholder-visual

171-placeholder-html

(no placeholder text for content in HTML mode, reflecting iOS)

Had some trouble with autocorrect (API 19+), the placeholder color would remain and color the actual text entered until space or entered was pressed. Fixed by setting a fixed color in the CSS and dropping the `ZSSEditor` methods that were handling setting it dynamically.

Note that the text in the editor is coming from the example app - the actual placeholders will be set from WPAndroid, so we can have 'Post Title' or 'Page Title' accordingly, as with iOS.

### DIFF
--- a/WordPressEditor/src/main/res/values/colors.xml
+++ b/WordPressEditor/src/main/res/values/colors.xml
@@ -10,7 +10,6 @@
     <color name="format_bar_button_normal_color">@color/wp_gray</color>
     <color name="format_bar_button_highlighted_color">@color/wp_blue</color>
 
-    <color name="placeholder_text">@color/wp_gray</color>
-
     <color name="sourceview_separator">@color/wp_gray_lighten_30</color>
+    <color name="sourceview_placeholder_text">@color/wp_gray</color>
 </resources>
--- a/WordPressEditor/src/main/res/values/colors.xml
+++ b/WordPressEditor/src/main/res/values/colors.xml
@@ -10,5 +10,7 @@
     <color name="format_bar_button_normal_color">@color/wp_gray</color>
     <color name="format_bar_button_highlighted_color">@color/wp_blue</color>
 
+    <color name="placeholder_text">@color/wp_gray</color>
+
     <color name="sourceview_separator">@color/wp_gray_lighten_30</color>
 </resources>
--- a/libs/editor-common/assets/editor.css
+++ b/libs/editor-common/assets/editor.css
@@ -287,6 +287,7 @@ div.field[contenteditable] {
 
 div.field[placeholderText][contenteditable=true]:empty:before {
     content: attr(placeholderText);
+    color: #87a6bc;
     transition: 0.2s ease opacity;
 }
 
--- a/libs/editor-common/assets/editor.css
+++ b/libs/editor-common/assets/editor.css
@@ -73,10 +73,10 @@ html, body {
 }
 
 body {
-    padding-left:5px;
-    padding-right:5px;
-    padding-top: 0px;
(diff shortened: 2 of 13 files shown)


aforcier:
cc @bummytime, @thianhlu

bummytime:
@aforcier In my testing so for, API 22 looks good, however API 16 (simulator) and 17 (simulator and device) had a few issues:

https://cloudup.com/clwCdpvn7mR
- The initial cursor position seems to be shifted left and up.
- The select all and cut behavior seems a little off. On new posts it seems to work fine, but on our existing test post the placeholder never seems to come back unless backspace is pressed multiple times.

libs/editor-common/assets/editor.css
```diff
@@ -287,6 +287,7 @@ div.field[contenteditable] {
 
 div.field[placeholderText][contenteditable=true]:empty:before {
     content: attr(placeholderText);
+    color: #87a6bc;
```
bummytime, line 289: A lot easier to do it here, eh? :-)

WordPressEditor/src/main/res/values/colors.xml
```diff
@@ -10,5 +10,7 @@
     <color name="format_bar_button_normal_color">@color/wp_gray</color>
     <color name="format_bar_button_highlighted_color">@color/wp_blue</color>
 
+    <color name="placeholder_text">@color/wp_gray</color>
```
bummytime: Just thinking out loud...do you think we need to be more specific with the name here since this value defines the placeholder color in the source view and not the visual view?

aforcier: Agreed, I'll rename that.

libs/editor-common/assets/editor.css
aforcier, line 289: One picks one's battles with Android autocorrect.

aforcier:
> The initial cursor position seems to be shifted left and up.

Nice catch, looks like it's indeed caused by setting a placeholder. Investigating.

> The select all and cut behavior seems a little off. On new posts it seems to work fine, but on our existing test post the placeholder never seems to come back unless backspace is pressed multiple times.

I'm able to reproduce this on an API 22 device. From inspection, it looks like Select All > Cut leaves behind some tags (usually a set of p tags, sometimes a `<br/>`), and so the field isn't understood to be empty by the CSS and the placeholder doesn't show. The wrapping tags show up when loading existing content and also when switching back and forth with HTML mode. Looking into a fix.

aforcier:
Updates:
- Fixed the cursor position bug.

It seems to be related to #179, as another oddity of empty fields for `API<19`. The contenteditable's padding CSS is ignored when there is placeholder text and the document falls back to the body CSS. I fixed this by moving the contenteditable padding to the body and the separator div. This solves the issue and doesn't affect the layout at all. However, focus is still lost if you delete the last character in the field, and you have to re-focus; this has the same cause as #179 and will be fixed at a later time.
- Fixed the Select All > Cut bug

Deleting all the contents of the document was sometimes leaving behind a carriage return character. The editor now checks if the field is empty except for a carriage return and empties it, re-displaying the placeholder text.
- Rename placeholder text color resource

Ready for another look @bummytime!

bummytime:
Ok @aforcier. Nice updates. :shipit:

PR #191, opened 2015-07-21
Issue #165: Content field not filling remaining screen space

### LINKED ISSUE #165: Keyboard disappears after tapping in editor
I can't repro this consistency, but sometimes when I tap in the editor after typing a title the keyboard disappears. [Video demo](https://cloudup.com/cUz88RZP-FE).

Note that tapping again in the editor doesn't cause the keyboard to re-appear, so there doesn't seem to be a way to type the post when this happens.

(Nexus 6,  Android 5.1)

aforcier:
Fixes #165. When the content field did not have enough text to fill the screen, tapping below the last line of text would not give focus to the field. This was especially irritating for empty documents.

(video of the issue: https://cloudup.com/cw8OQOfyTVd)

Fixed by calling the existing `ZSSEditor.refreshVisibleViewportSize()` whenever the `WebView`'s layout changed (initial setup, keyboard appearing and hiding).

cc @bummytime

### DIFF
--- a/WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragment.java
+++ b/WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragment.java
@@ -94,6 +94,20 @@ public class EditorFragment extends EditorFragmentAbstract implements View.OnCli
         mWebView.setOnTouchListener(this);
         mWebView.setOnImeBackListener(this);
 
+        // Ensure that the content field is always filling the remaining screen space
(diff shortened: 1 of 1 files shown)


bummytime:
Heh, we had this same issue on iOS.

bummytime:
Could not repro issue at all. Also checked with the device inspector. Nice job @aforcier.

Tested on:
- API 16 :+1: 
- API 18 :+1: 
- API 19 :+1: 
- API 21 :+1:

Unit tests: Green
Connected Tests: Green

PR #192, opened 2015-07-21
Update to new containers

kwonye:
http://docs.travis-ci.com/user/workers/container-based-infrastructure/

Needs review: @aforcier

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: android
 jdk: oraclejdk7
+sudo: false
 
 android:
   components:


aforcier:
Looks good @kwonye. Since we set up Travis in April 2015 it looks like we were already using containers, but it makes sense to make it explicit.

:shipit:

kwonye:
Thanks for the quick review!

> Since we set up Travis in April 2015 it looks like we were already using containers

Interesting! I wrongly assumed travis was set up when the repo was initialized. I actually don't mind reverting this then considering containers are now a default and `sudo: false` is for "legacy" projects. It's up to you.

aforcier:
Hmm, looks like you're right, it's only meant for 'legacy':

http://docs.travis-ci.com/user/migrating-from-legacy/#How-can-I-use-container-based-infrastructure%3F

I'll revert this then :+1:

PR #195, opened 2015-07-21
Fixes hiding/showing of action bar when changing orientations

### LINKED ISSUE #173: Inconsistent ActionBar behavior on rotation
When not recreating the fragment on rotation (e.g., from within the WPAndroid project), the ActionBar doesn't always behave properly.

Steps to recreate:
- Enter landscape mode on a device smaller than a large tablet
- Tap a field, causing the soft keyboard to appear and the ActionBar to hide
- Rotate the device
- The keyboard remains up, but the ActionBar remains hidden, when it should be visible at all times in portrait mode

The converse also happens with the ActionBar staying visible in landscape mode after rotating from portrait, when it should be hidden.

This is probably fixable by forcing an ActionBar show/hide inside `onConfigurationChanged`.

bummytime:
Fixes #173.

To test, run wpandroid with the updated editor branch:

`git subtree pull --prefix=libs/editor git@github.com:wordpress-mobile/WordPress-Editor-Android.git issue/173-inconsistent-actionbar`

Try it out on a variety of APIs as well as a large tablet.

/cc @aforcier

### DIFF
--- a/WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragment.java
+++ b/WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragment.java
@@ -203,10 +203,10 @@ public class EditorFragment extends EditorFragmentAbstract implements View.OnCli
         if (newConfig.orientation == Configuration.ORIENTATION_LANDSCAPE
                 && !getResources().getBoolean(R.bool.is_large_tablet_landscape)) {
             mHideActionBarOnSoftKeyboardUp = true;
-            this.hideActionBarIfNeeded();
+            hideActionBarIfNeeded();
         } else {
             mHideActionBarOnSoftKeyboardUp = false;
-            this.showActionBarIfNeeded();
+            showActionBarIfNeeded();
         }
     }
 
@@ -275,7 +275,7 @@ public class EditorFragment extends EditorFragmentAbstract implements View.OnCli
         if (event.getAction() == MotionEvent.ACTION_UP) {
             // If the WebView or EditText has received a touch event, the keyboard will be displayed and the action bar
             // should hide
-            this.hideActionBarIfNeeded();
+            hideActionBarIfNeeded();
         }
         return false;
     }
@@ -285,7 +285,7 @@ public class EditorFragment extends EditorFragmentAbstract implements View.OnCli
      */
     @Override
     public void onImeBack() {
-        this.showActionBarIfNeeded();
+        showActionBarIfNeeded();
     }
 
     @SuppressLint("NewApi")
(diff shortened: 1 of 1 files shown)


aforcier:
Noticed one remaining `ActionBar` glitch that wasn't described in the original issue:
- start in portrait and dismiss the soft keyboard
- rotate to landscape

The keyboard stays hidden, but the ActionBar is also hidden. Bringing up the keyboard will keep the ActionBar hidden, and hiding the keyboard again will correctly show the ActionBar.

This also happens if already in landscape mode when the editor is launched. The behaviour is the same with a hardware keyboard attached.

This is particularly troublesome if you're using a hardware keyboard and you disable the software keyboard, as that leaves you no way of showing the ActionBar (and therefore publishing) while in landscape mode.

Otherwise, the fix is working well. Tested on:

API 14 :+1:
API 15 :+1:
API 16 :+1:
API 17 :+1:
API 18 :+1: (including Samsung S3)
API 19 :+1:
API 21 :+1:
API 22 :+1: (including Nexus 4 & Nvidia Shield [small tablet] with hardware keyboard)
API M :+1: (Nexus 9 [small tablet], Android M Prev 2)

Also tested on an API 22 Nexus 10 emulator [large tablet], ActionBar never hides, as expected.

(the issue mentioned above appears on all of these devices as well)

Up to you if you want to bake a fix for the issue above into this or cut a separate issue, @bummytime.

bummytime:
> The keyboard stays hidden, but the ActionBar is also hidden. Bringing up the keyboard will keep the ActionBar hidden, and hiding the keyboard again will correctly show the ActionBar.

Nice catch @aforcier. Lemme take a look at it.

bummytime:
@aforcier Not sure if it is 100% there, but try testing my latest commit with wpandroid. If it is :hankey:, I can revert it and open a separate issue.

bummytime:
Also, try testing with devices if possible. :grinning:

aforcier:
Looks like the fix works, but only when using a hardware keyboard with the software keyboard disabled (which was in fact the worst case scenario), with one exception (see Samsung S3 below).

In HW keyboard + disabled soft keyboard cases, the ActionBar is always displayed. The fix doesn't break anything, but the issue is still present if you're using a software keyboard (the ActionBar will hide when flipping to landscape with the keyboard hidden).

Devices tested:

API 14 emulator: works for HW keyboard

API 18 emulator: works for HW keyboard

API 18 device: still has the bug, even with HW keyboard connected and soft keyboard disabled (Samsung S3) - likely a Samsung keyboard issue

API 22 devices: works for HW keyboard (tested on both Nexus 4 and Shield with/without HW keyboard)

API M device: same as 22 (Nexus 9)

Also checked a large tablet emulator (API 22) - still always displaying the ActionBar as expected.

aforcier:
To clarify, using a software keyboard with no HW keyboard at all will also reproduce the new rotation bug on all the devices listed.

aforcier:
With the initial issue and the worst case fixed, I think this is a solid improvement for landscape mode. As discussed, I think we should open a separate issue if the last bug doesn't end up having a straightforward solution. Given that and the style nit I mentioned, :shipit:.

bummytime:
Ok, giving up on that last bug for now. Filed issue #196 to track it.

PR #205, opened 2015-07-29
Allow WordPress formatting to be disabled

alexcrist:
Hey there,

First of all, I want to say a big thanks for a great library! I've found this to be a useful and full featured HTML editor that beats the others I've looked at.

I would love to implement the editor in a project I'm working on, but I need to keep `<p>` tags from being stripped out. I do understand that keeping the Android editor's formatting consistent with the web app is important.

---

My proposed change is to create a new property of EditorFragment: `mWordPressFormatting`, a `boolean` variable set to `true` by default. 
- When set to `true`: the editor functions normally. 
- When set to `false`: the editor skips formatting the html content according to the rules in `wpload.js` and `wpsave.js`.

A new method is implemented into `EditorFragment`, `setWordPressFormatting(boolean enabled)` allows toggling of `mWordPressFormatting`.

---

I would love to hear what you guys think about this!

Thanks again,
Alex Crist

### DIFF
--- a/WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragment.java
+++ b/WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragment.java
@@ -61,6 +61,8 @@ public class EditorFragment extends EditorFragmentAbstract implements View.OnCli
     private CountDownLatch mGetTitleCountDownLatch;
     private CountDownLatch mGetContentCountDownLatch;
 
+    private boolean mWordPressFormatting = true;
+
     private final Map<String, ToggleButton> mTagToggleButtonMap = new HashMap<>();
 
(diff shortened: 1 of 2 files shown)


aforcier:
Hey there,

Thanks for the suggestion!

We consider this library incomplete right now, work is in progress and we have even more WordPress-specific stuff to add in the next few months. Since we don't know yet what the full scope of WordPress-specific settings will be (and there are already a bunch of things besides `p`-tag stripping one might want to turn off for a separate project), we're not comfortable adding this property just yet.

This is something we'd like to revisit when the library is built up and we can extract what WP-only features we can in a methodical way. In the meantime working off a fork might be the simplest approach.

alexcrist:
Ah, that makes sense, thanks for the info!

PR #209, opened 2015-07-29
Issue #203: Restore legacy editor's original format bar size

### LINKED ISSUE #203: Set legacy format bar height back to 40dp
Title has it.

aforcier:
Fixes #203.

We've been using a shared resource for the format bar height for the new and legacy editors. A recent change increased the new editor's format bar size from `40dp` to `44dp`, and that changed the legacy editor as well.

This change gives the legacy editor its own format bar height dimen and sets it back to `40dp`.

Note: this did not affect WPAndroid, as the WPAndroid `develop` branch currently overwrites `format_bar_height` and sets it to `40dp` anyway.

cc @bummytime


bummytime:
:shipit:

PR #220, opened 2015-08-07
Issue #197: Dimensions not updated on rotation

### LINKED ISSUE #197: Dimensions not updated on rotation (WPAndroid)
In WPAndroid, since the editor wrapping activity is not re-created on rotation (to prevent the keyboard from hiding), screen size-specific layouts are not recalculated when rotating.

For tablets, this means that the side margins for HTML mode are fixed to either landscape or portrait mode (whichever mode the editor was launched in). This also means that launching in landscape mode on a small tablet and then rotating to portrait will hide the last few icons in the format bar (as we have custom dimensions for the spacing between them):

wpandroid-tablet-rotation-bug

(landscape-sized margins for HTML mode and hidden blockquotes and HTML buttons)

This doesn't affect phones, or any device square enough to land in the same sizing bucket in portrait and in landscape.

aforcier:
Fixes #197 by reloading parts of the editor layout on rotation.

`onConfigurationChanged()` now removes and re-attaches the format bar to the view, so that the correct format bar size is shown. It does the same for the side margins in HTML mode.

This PR also changes the example activity to preserve the editor fragment on configuration change (just like WPAndroid).

cc @bummytime


bummytime:
@aforcier You may want to update this branch from `develop`

aforcier:
@bummytime updated!

bummytime:
Tested on:
- API 16 :+1: 
- API 18 :+1: 
- API 19 :+1: 
- API 21 :+1:

Unit tests: Green
Connected Tests: Green

Reviewing code now....

WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragment.java
```diff
@@ -218,6 +237,42 @@ public void onConfigurationChanged(Configuration newConfig) {
         }
     }
 
+    private void setupFormatBarButtonMap(View view) {
```
bummytime, line 240: Nice job busting this out into it's own method. :+1:

bummytime:
Good stuff @aforcier. I like the format bar having it's own layout...seems logical. Code looks good, nothing really sticks out to me. :shipit:

PR #221, opened 2015-08-07
Issue #218: HTML mode crash on API 16

### LINKED ISSUE #218: HTML mode can crash on API 16
In API 16 versions `4.1` and `4.1.1` only (fixed in `4.1.2`), switching to HTML mode can cause the app to crash. This is due to an Android bug - the crash is triggered by having spanned text that is being line-wrapped. Here is the AOSP issue:

https://code.google.com/p/android/issues/detail?id=35466

This happens consistently with the example app on a Nexus 4 emulator running `4.1.1`. In actual use of the editor, the crash would depend on the device size and the HTML contained in the post being edited. If any one piece of formatted text has to line-wrap, the crash occurs on loading HTML mode.

The simplest fix would be to disable coloring the HTML for those versions (functionally it would all work the same as on other versions, we just wouldn't color the HTML tags).

This would affect under 1% of users (~6% are using `API 16`, but most of those are on the patched `4.1.2` version).

aforcier:
Fixes #218. Added a check which skips applying spans to HTML mode contents when on Android `4.1` and `4.1.1`, to avoid a crashing bug in those versions of Android.

cc @bummytime

### DIFF
--- a/WordPressEditor/src/main/java/org/wordpress/android/editor/HtmlStyleUtils.java
+++ b/WordPressEditor/src/main/java/org/wordpress/android/editor/HtmlStyleUtils.java
@@ -2,6 +2,7 @@ package org.wordpress.android.editor;
 
 import android.graphics.Color;
 import android.graphics.Typeface;
+import android.os.Build;
 import android.support.annotation.NonNull;
 import android.text.Spannable;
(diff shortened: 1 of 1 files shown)


bummytime:
Seems to work.

shrug

Nice work. :shipit:

Issue #210, opened 2015-07-29
Blockquote after list wraps list in quote


meremagee:
Adding a blockquote after lists wraps lists in quote as well.

device: SM-T330NU (samsung)
Android version: 4.4.2

repro:
- add a list (ol or ul)
- exit list
- type regular p text
- tap `blockquote` to quote p text
- note that previous list items are also wrapped in blockquote

video: https://cloudup.com/cOgZJgJQfDO

![](https://cloudup.com/iCSjF6Njv-X+)

_possibly related, though I cannot repro https://github.com/wordpress-mobile/WordPress-Editor-Android/issues/161_

bummytime:
While in the visual editor if you tap the list format bar button (when you are all done with your list) the editor does not create a new paragraph. Instead it puts the new text directly under the existing list: 

![screen shot 2015-08-12 at 7 09 37 pm](https://cloud.githubusercontent.com/assets/154014/9239905/0e41164c-4126-11e5-93dc-99cc66113d42.png)
When the blockquote button is pressed, it surrounds the parent node of the node where the cursor is with a `<blockquote>` tag:

![screen shot 2015-08-12 at 7 20 22 pm](https://cloud.githubusercontent.com/assets/154014/9239999/43d2d524-4127-11e5-8805-801f63b1ef4f.png)
This is "technically" the correct behavior of our blockquote logic. The question is when a list is ended, do we automagically create a new `<p>`? When this is done, the blockquote logic works as expected. 

It is interesting to note that if you take the post content above (in the 1st image) and pop into HTML mode and then back again, our load/unload scripts actually "fix" the HTML:

![screen shot 2015-08-12 at 7 10 07 pm](https://cloud.githubusercontent.com/assets/154014/9240162/b7242c70-4128-11e5-9517-b383df88d8d9.png)

/cc @meremagee

bummytime:
Also, wp-admin seems to do this:

![screen shot 2015-08-12 at 9 15 44 pm](https://cloud.githubusercontent.com/assets/154014/9241299/5ca5d30c-4137-11e5-931a-f744a7121f8a.png)

PR #226, opened 2015-08-14
Fix for Blockquote wrapping sibling list

bummytime:
This PR takes a crack at Fixing #210. Here is a video of it working: 

https://cloudup.com/c0HWu_hIJ9q. 

I had to do some gymnastics to fix how lists are inserted via `execCommand`.

@aforcier would you mind starting to take a look at this. Mainly I need you to help test it out on a variety of devices and API's. Basically just test adding lists and bquotes in a bunch of different ways.

There is a known "issue" right now. When turning off a list and the caret is after a word already in a `<li>`, blockquoteing that word cause the list to be blockquoted as well. This video demonstrates it better than I can describe:

https://cloudup.com/c2aW80mp-0L

Not sure if this issue is REALLY a high priority issue, nevertheless I think this fix helps improve matters a tad.

/cc @meremagee

### DIFF
--- a/libs/editor-common/assets/ZSSRichTextEditor.js
+++ b/libs/editor-common/assets/ZSSRichTextEditor.js
@@ -25,7 +25,8 @@ const NodeName = {
     UL: "UL",
     LI: "LI",
     CODE: "CODE",
-    SPAN: "SPAN"
+    SPAN: "SPAN",
+    BR: "BR"
 };
 
 // The editor object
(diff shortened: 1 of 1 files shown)


meremagee:
Can we test closing list by hitting enter/return 2x in addition to closing by tapping active list icon?

bummytime:
@meremagee I forgot to mention that, I am working on fixing that still.

/cc @aforcier

aforcier:
Found an issue in `API < 19` when entering a blockquote after a list, if the list was the first thing added to an empty post.

Here's the behavior on `develop`:

https://cloudup.com/cZTSzDJ_jV4

And here's the behavior on this branch:

https://cloudup.com/cOIcHZM5Cm9

This is accompanied by many instances of this error in the log (they start coming as soon as the list button is pressed to end the list):

`Uncaught TypeError: Cannot call method 'toLowerCase' of undefined -- From line 2017 of file:///android_asset/ZSSRichTextEditor.js`

The bug doesn't happen (and the above error isn't printed) if there was some other text entered before the list was added.

Extra note: This not only doesn't happen in `API >= 19`, the original comment issue doesn't seem to happen either: https://cloudup.com/cmc5pMNW_Rz

aforcier:
I'm also seeing the `'toLowerCase'` error on `API < 19` when hitting enter twice to end a list (as @meremagee brought up). There's a difference between API < 19 and API >= 19 here.

On `API >= 19`, the same thing happens as demonstrated in the screen capture in the original comment (https://cloudup.com/c2aW80mp-0L). I'm guessing that this is what you mentioned you're working on fixing @bummytime (with respect to double-enter).

On `API < 19`, nothing happens at all - the blockquote button turns on and off when tapping but no blockquotes are added or removed.

This bug, however, wasn't introduced by this PR - I'm able to reproduce it in `develop`, with the same error from the same line of code (though the line number is different of course). It seems to happen because double-enter doesn't clear the highlight on the list button on those APIs:

https://cloudup.com/cl3_DMSk1hd

This should probably be a new issue once we've cleared up what to do with the rest of the issues.

aforcier:
Apart from the issues mentioned above, the actual fix seems to work properly everywhere, and the issue demonstrated in the original comment is happening consistently everywhere as well.

Tested on emulators:
API 14: :+1:
API 15: :+1:
API 16: :+1:
API 17: :+1:
API 18: :+1: (also Samsung S3)
API 19: :+1:
API 21: :+1:
API 22: :+1: (also Nexus 4 and Nexus 9)

bummytime:
Thanks for looking @aforcier. I am working on cleaning up those other issues.

aforcier:
Added a note and screen capture at the bottom of my first comment showing a case where the first comment's issue doesn't happen.

bummytime:
@aforcier Ok. I added some logic to handle the double enter situation @meremagee mentioned above. Would you mind giving it a try?

aforcier:
Tested the double enter fix on APIs `14`-`22`, all good :+1:.

bummytime:
Ok @aforcier I broke out those two bugs separately. Feel free to merge if you are cool with things.

aforcier:
Looks great @bummytime! :shipit:

PR #229, opened 2015-08-15
Fix for Format bar: html cut off (Samsung SM-T) #219

### LINKED ISSUE #219: Format bar: html cut off (Samsung SM-T)
In both disabled and active modes, the HTML button is only partially visible on the Samsung SM-T (android 4.4.2) in portrait mode. It fully displays on landscape.

https://cloudup.com/ceNlvS3sufG

bummytime:
Fixes #219.

Made the minimum width of the tablet layout 610dp.

/cc @aforcier


bummytime:
Based off of offline discussion regarding format bar layout on the Galaxy Tab 4:

screenshot_2015-08-12-09-55-18

601px. :-(

aforcier:
Throwing this out there - what if we kept the minimum width for a tablet at `600dp` and dropped a bit of padding instead? I think it would be good to stay aligned with the Android/Material style guide's definition of a tablet, if we can:

https://www.google.com/design/spec/layout/adaptive-ui.html#adaptive-ui-breakpoints

aforcier:
@thianhlu what do you think?

thianhlu:
How's it look? We were going to remove the padding anyways right?

aforcier:
The plot thickens, here's how `develop` looks on a 2013 Nexus 7 (also `600dp` wide):

229-nexus7-portrait

Screen size information:

229-nexus7-screen-size

Not sure what's happening here, as it should also be cutting off part of the HTML button just like the Galaxy Tab 4. Will revisit this after @meremagee tests the 0.3 build on her Galaxy Tab 4 (in case the issue was actually related to #220 after all).

aforcier:
From searching, it looks like the Samsung Galaxy Tabs have a history of mis-reporting their density. I didn't find anything specific to the Galaxy Tab 4, but it looks like it might be using `xhdpi` resources when it should be using `hdpi`.

I sent @meremagee an apk of the example app with the padding here dropped to `0dp`:

https://github.com/wordpress-mobile/WordPress-Editor-Android/blob/244a51315e24eddc0510b422a75e6483c6c0f6f3/WordPressEditor/src/main/res/values/dimens.xml#L11

I'd suggest dropping as much padding as needed for everything to look right on the SM-T. Then we can take a look at the Nexus 7 which uses the correct resources at `600dp` and decide if it still looks good with some dropped padding.

bummytime:
@meremagee did you have any time to check out that APK above?

meremagee:
@bummytime @aforcier yep! sorry for not posting sooner. HTML icon looks good on that build: https://cloudup.com/iTz__mbPy7R

aforcier:
Here's what the same build (0dp padding between icons, padding between groups untouched) looks like on a Nexus 7 (2013) (real 600dp width):

226-format-bar-0dp-padding-nexus7

cc @thianhlu

thianhlu:
:+1:

bummytime:
@aforcier did you have code from your APK you want to update this PR with or should I do it?

aforcier:
@bummytime I'll push up my changes momentarily

aforcier:
:shipit:

Issue #2, opened 2015-03-25
New post workflow (WIP)


bummytime:
Using the new visual editor on iOS as inspiration:

![screen shot 2015-03-25 at 10 32 44 am](https://cloud.githubusercontent.com/assets/154014/6829061/24b44972-d2df-11e4-8101-563023b44d43.png)

 the Android editor will be updated to reflect the new visual look & feel as well as have 100% visual editing for new _and_ existing posts. For context, here are the existing editor screens on Android:

<h4>New Post: </h4>

![new_post_1](https://cloud.githubusercontent.com/assets/154014/6829613/888eb074-d2e2-11e4-858d-d743e23931a2.png)

<h4>New Post Title:</h4>

![new_post_title](https://cloud.githubusercontent.com/assets/154014/6829678/feb2e248-d2e2-11e4-9113-a4905fe54238.png)

<h4>New Post Content:</h4>

![new_post_content](https://cloud.githubusercontent.com/assets/154014/6829688/1278fbfa-d2e3-11e4-8d7b-b31c95fa76aa.png)

@thianhlu We'll need you to provide some input here. :grinning:

bummytime:
[Here's a video](https://cloudup.com/cCNCgX9Hme2) of the existing new/edit post workflow for posterity's sake.

aforcier:
Closing - the new Android editor now closely matches the iOS editor's new post workflow:
- Placeholders in empty fields
- Text styling, side margins
- Post settings button in ActionBar rather than bottom button
- No more full-screen content editing mode

Issue #187, opened 2015-07-16
Insert media into post via visual editor


bummytime:
In the current Android editor, when the user taps the media icon on the format bar the are presented with this:

![2015-07-16 12 31 09](https://cloud.githubusercontent.com/assets/154014/8730546/0a96f018-2bb7-11e5-86c8-5b7e74948e50.png)

and selecting one of those options presents UI where the user can select/create media and insert it into the post. Maybe we just use this existing flow?  

While media is uploading it would be great to show a progress indicator and slowly raise the opacity of the image in the post (from iOS):

![screen shot 2015-07-16 at 12 38 31 pm](https://cloud.githubusercontent.com/assets/154014/8730626/9af0e916-2bb7-11e5-9e2f-14bb4078d635.png)

and if the image upload fails show a failed overlay (with retry option). See: https://github.com/wordpress-mobile/WordPress-Editor-iOS/issues/361

![screen shot 2015-07-16 at 12 49 15 pm](https://cloud.githubusercontent.com/assets/154014/8730851/1fcb8b04-2bb9-11e5-8b0c-82abe895fc88.png)

some of this UI/UX is already built into the JS/HTML in iOS so limited re-use should be possible.

Issue #13, opened 2015-03-30
Edit Images


aforcier:
Tapping an image loads the image settings screen, which we might want to change to mirror the options offered in the iOS version. Here are the current screens on Android and iOS:

![01-image_properties_comparison](https://cloud.githubusercontent.com/assets/9613966/6908073/f5a7e6e2-d708-11e4-95f1-ab7ab6485427.png)

We might also want to change the tap interaction itself. Currently in Android, tapping an image while in edit mode will launch the 'Image settings' screen right away. By contrast, the iOS editor requires two taps. The first blurs the image and overlays an 'Edit' message:

![02-ios_edit_image](https://cloud.githubusercontent.com/assets/9613966/6908076/f98ea1ba-d708-11e4-98a0-5b880e52ce97.png)

Tapping again will display the 'Edit image' screen for iOS shown above.

bummytime:
Just as a FYI, this is the current image edit overlay:

![screen shot 2015-03-30 at 6 41 02 pm](https://cloud.githubusercontent.com/assets/154014/6909523/a5c117ee-d70c-11e4-8a35-d22b9a5d6e38.png)

> We might also want to change the tap interaction itself. Currently in Android, tapping an image while in edit mode will launch the 'Image settings' screen right away. By contrast, the iOS editor requires two taps. 

I am not really sure about the genesis of the two tap gesture on iOS. Perhaps @thianhlu knows?

Also Ref: https://github.com/wordpress-mobile/WordPress-Editor-iOS/issues/549 and https://github.com/wordpress-mobile/WordPress-Editor-iOS/issues/427

aforcier:
I'm re-working the image setting screen and need some input. Here's the current version (needs some polish like aligning the captions with the EditTexts):

![13-image-settings-sample](https://cloud.githubusercontent.com/assets/9613966/9600950/f9e3e75c-50a6-11e5-9f20-eedae362cea6.png)

A couple of points:
- I added the missing options from iOS (alt text and link)
- I dropped the selection bar for width, replacing it with a spinner with the options iOS has (Full Size, Thumbnail, Medium and Large)
- Especially with the alt text field now added, I think that labels are necessary for the 3 text fields at the top. If they're all filled out, there are no hints, and it's not at all clear which is the title, the caption and the alt text. I used the existing labels to demo it out.
[...]

PR #236, opened 2015-08-26
Feature #187: Insert media

aforcier:
Addresses #187. Adds basic support for inserting images into the visual editor. Doesn't include editing image settings - that will be addressed in a separate PR (issue #13).

This PR also adds some demo functionality to the example app, to simulate handling uploads and making appropriate calls back to the editor fragment. Integration with WPAndroid will be introduced in a future PR.

187-media-uploading

Tapping an uploading image shows a prompt allowing the user to cancel the upload:

187-media-stop-upload-dialog

An overlay is shown when an upload fails. Tapping the image in this state restarts the upload:

187-media-upload-fail-overlay

HTML mode is disabled during upload

187-media-upload-html-toast

<h2>Known issues:</h2>

<h3>1: Samsung S3 image spacing issue</h3>

There is currently a bug occurring on the Samsung S3 (`API 18`). It ignores the space and you're stuck with the cursor on the left side of the image - no way to keep typing under the image without going into HTML mode. This didn't happen on emulators (`API 14-22`), but I trust physical devices more - we might need some testing on older devices to figure out the extent of the issue.

The S3 issue can be fixed by using a zero-width space character (`U+200B`) instead of `<br>` as a separator [here](https://github.com/wordpress-mobile/WordPress-Editor-Android/blob/ddb5454bd7a6a932eaf0bbb5ea0ddb531a16233b/libs/editor-common/assets/ZSSRichTextEditor.js#L834). This works on all devices, but messes up autocorrect - the ZW space character is treated as a real character, so the keyboard switches to lowercase under the image (as if a sentence had already started). There might be a different solution for the spacing that will work everywhere with no issues, or we might be able to force the autocapitalization back on after images.

<h3>2: Missing upload progress bar (API < 19)</h3>

On all `API < 19` devices, the progress bar doesn't show up when uploading images. It seems that Android didn't support the `<progress>` tag we're using until the Chromium WebView (`API 19`).  One possibility here is to use an upload notification instead. The WPAndroid app uses those for post uploads and gallery uploads, so we might want to use it for individual image uploads anyway (on all APIs, even ones where the progress bar inside the WebView is shown). Here's a demo of what that would look like on an API 18 device (the notification isn't included in this PR, just looking for feedback):

https://cloudup.com/cqzRAsg_CeE

<h3>3: Uploading image container is sometimes very small (API < 19)</h3>

Sometimes (probably when uploading media in a blank document, still testing) the image container is very small while uploading. Once the container is removed at the end of the upload, the normal-sized image is shown:

https://cloudup.com/c2HhsInzXsW

The problem is resolved by removing [this](https://github.com/wordpress-mobile/WordPress-Editor-Android/blob/244a51315e24eddc0510b422a75e6483c6c0f6f3/libs/editor-common/assets/editor.css#L141) line from the CSS. However, this probably isn't the root cause, and it also breaks the progress bar on `API > 18`.

If we can't fix issue #2 above, we might as well drop the container entirely (and load a different CSS file for API > 18`and for`API <= 18`).

<h3>4: Bugginess with the edit overlay on some API levels</h3>

I'll address this one as part of the future PR for #13.

### DIFF
--- a/WordPressEditor/src/androidTest/java/org.wordpress.android.editor/MockEditorActivity.java
+++ b/WordPressEditor/src/androidTest/java/org.wordpress.android.editor/MockEditorActivity.java
@@ -47,6 +47,11 @@ public class MockEditorActivity extends AppCompatActivity implements EditorFragm
     }
 
     @Override
+    public void onMediaRetryClicked(String mediaId) {
+
+    }
+
+    @Override
     public void saveMediaFile(MediaFile mediaFile) {
 
     }
--- a/WordPressEditor/src/androidTest/java/org.wordpress.android.editor/MockEditorActivity.java
+++ b/WordPressEditor/src/androidTest/java/org.wordpress.android.editor/MockEditorActivity.java
@@ -52,6 +52,11 @@ public class MockEditorActivity extends AppCompatActivity implements EditorFragm
     }
 
     @Override
+    public void onMediaUploadCancelClicked(String mediaId) {
+
+    }
+
+    @Override
     public void saveMediaFile(MediaFile mediaFile) {
 
     }
--- a/WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragment.java
+++ b/WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragment.java
@@ -262,6 +262,8 @@ public class EditorFragment extends EditorFragmentAbstract implements View.OnCli
         ToggleButton mediaButton = (ToggleButton) view.findViewById(R.id.format_bar_button_media);
         mTagToggleButtonMap.put(TAG_FORMAT_BAR_BUTTON_MEDIA, mediaButton);
 
+        registerForContextMenu(mediaButton);
+
         ToggleButton linkButton = (ToggleButton) view.findViewById(R.id.format_bar_button_link);
         mTagToggleButtonMap.put(TAG_FORMAT_BAR_BUTTON_LINK, linkButton);
 
@@ -323,6 +325,11 @@ public class EditorFragment extends EditorFragmentAbstract implements View.OnCli
         } else if (id == R.id.format_bar_button_media) {
             // TODO: Handle inserting media
             ((ToggleButton) v).setChecked(false);
+
+            mEditorFragmentListener.onAddMediaClicked();
+            if (isAdded()) {
+                getActivity().openContextMenu(mTagToggleButtonMap.get(TAG_FORMAT_BAR_BUTTON_MEDIA));
+            }
         } else if (id == R.id.format_bar_button_link) {
             if (!((ToggleButton) v).isChecked()) {
                 // The link button was checked when it was pressed; remove the current link
--- a/WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragment.java
+++ b/WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragment.java
@@ -547,15 +547,15 @@ public class EditorFragment extends EditorFragmentAbstract implements View.OnCli
     }
 
     @Override
-    public void appendMediaFile(final MediaFile mediaFile, final String imageUrl, ImageLoader imageLoader) {
+    public void appendMediaFile(final MediaFile mediaFile, final String mediaUrl, ImageLoader imageLoader) {
         mWebView.post(new Runnable() {
             @Override
             public void run() {
-                if (URLUtil.isNetworkUrl(imageUrl)) {
-                    mWebView.execJavaScriptFromString("ZSSEditor.insertImage('" + imageUrl + "');");
+                if (URLUtil.isNetworkUrl(mediaUrl)) {
+                    mWebView.execJavaScriptFromString("ZSSEditor.insertImage('" + mediaUrl + "');");
                 } else {
                     String id = mediaFile.getMediaId();
-                    mWebView.execJavaScriptFromString("ZSSEditor.insertLocalImage(" + id + ", '" + imageUrl + "');");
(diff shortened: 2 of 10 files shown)


aforcier:
cc @bummytime @thianhlu

thianhlu:
The visuals look great @aforcier, impressive! 

Question on the Samsung issue ^: Once the image upload is complete, does the spacing issue resolve itself?

bummytime:
Just a general comment on something I noticed while testing on the Nexus4 (API 17). It seems like occasionally, backspacing over (aka deleting) an image (usually a failed upload) will cause strange behaviors in the editor when trying to edit or add a new line around the area of the image that was deleted.

For example, in one instance I backspaced/deleted an image and then hit return — every time I hit return a `<li>` appeared....and there was not any lists in the post.

Another time after deleting a photo, every subsequent time I hit backspace caused a new line to be inserted.

I also noticed some of these errors flying around:

`Uncaught Error: INDEX_SIZE_ERR: DOM Exception 1 -- From line 1991 of file:///android_asset/ZSSRichTextEditor.js`

`Uncaught Error: INDEX_SIZE_ERR: DOM Exception 1 -- From line 2040 of file:///android_asset/ZSSRichTextEditor.js`

which makes sense given that both of these functions try to return the parent node and cannot hence the strange behavior.

Naturally I cannot repro these issues consistently. Just wanted to make a note of it here.

bummytime:
Also @aforcier I noticed occasionally selected photos that are not on device (e.g. Google Photos or Picassa) will cause the picker to return and nothing is inserted into the post at all. This happened about 5% of the time for me on devices < API 19.

bummytime:
In the iOS version, we are able to cancel the an image upload by tapping an in-progress upload:

scaled screen shot 2015-08-26 at 3 04 20 pm

@meremagee @aforcier Are we looking to do a similar feature here? (I'm cool with it being a separate issue if need be).

thianhlu:
Isn't that second example/image above?

bummytime:
@thianhlu HAH! yes. However it is not working for me...investigating.

bummytime:
@thianhlu @aforcier Oooooook. Ignore my comment above. The stop upload is working, however I needed to tap twice (once for webview focus and once to trigger the dialog).

bummytime:
@aforcier Looks like video is not supported currently (I am unable to select any videos). We could make a separate issue for this if desired.

bummytime:
Tested on:
- API 16 :+1: 
- API 17 :+1:  (including Nexus4 device) Occasional issues with removing photos noted above.
- API 18 :+1: 
- API 19 :+1: 
- API 21 :+1:
- API 22 :+1:  (including Samsung S6 device and Nexus7 2nd gen)

Unit tests: Green
Connected Tests: Green

bummytime:
Will review code shortly...

aforcier:
> Question on the Samsung issue ^: Once the image upload is complete, does the spacing issue resolve itself?

Unfortunately not, the space issue remains.

> Also @aforcier I noticed occasionally selected photos that are not on device (e.g. Google Photos or Picassa) will cause the picker to return and nothing is inserted into the post at all. This happened about 5% of the time for me on devices < API 19.

Since WPAndroid has a lot of media handling code between the picker returning and sending a URI to the editor, it's likely that this issue might not exist after we integrate media. So I suggest that we ignore this for now and look for this issue in the future WPAndroid integration PR.

> Looks like video is not supported currently (I am unable to select any videos). We could make a separate issue for this if desired.

Ah yes, I meant to note that. I didn't add video or gallery support - I'll open up separate issues for them.

bummytime:
@aforcier Code looks square.

> Since WPAndroid has a lot of media handling code between the picker returning and sending a URI to the editor, it's likely that this issue might not exist after we integrate media. So I suggest that we ignore this for now and look for this issue in the future WPAndroid integration PR.

:+1: :hamster: :sake: :santa: 

> The problem is resolved by removing this line from the CSS. However, this probably isn't the root cause, and it also breaks the progress bar on API > 18.
> 
> If we can't fix issue #2 above, we might as well drop the container entirely (and load a different CSS file for API > 18and forAPI <= 18`).

My vote would be to load a different CSS file for these lower APIs. The "small image" issue happened frequently enough in my testing that we should address it.

> On all API < 19 devices, the progress bar doesn't show up when uploading images. It seems that Android didn't support the <progress> tag we're using until the Chromium WebView (API 19). One possibility here is to use an upload notification instead. 

I like the idea of the notification on upload completion however I am not sure if that is enough. In testing, I found myself looking for more immediate feedback after I selected the image to upload. Instead I saw a transparent image for a few seconds and then a fully opaque image popped up (on 3G/4G this will probably take a lot longer). 

Just thinking out loud, I am wondering if we could do something else for these lower APIs...like either an indeterminate spinner overlay or even an overlay that just says "Uploading..." or something. @thianhlu and @meremagee what say you?

thianhlu:
@bummytime would the spinner be blocking for these cases?

bummytime:
@thianhlu Nah. I was just thinking about putting something on the image itself to let the user know it's actively being uploaded. The user could still write text or do other things during this time.

bummytime:
@aforcier I created separate issues for [issue 1](https://github.com/wordpress-mobile/WordPress-Editor-Android/issues/241) and [issue 2 and 3](https://github.com/wordpress-mobile/WordPress-Editor-Android/issues/242) above.

Given that we isolated these things from this PR, I am kewl with merging this PR if you are. Let me know your thoughts...

aforcier:
Conflicts resolved. I concur, I think we can merge this in now and attack the issues later on, since they're fairly isolated from the rest of the media changes (image settings).

bummytime:
:shipit:

Issue #112, opened 2015-04-04
Use image resize dimensions in blog's options


bummytime:
Original issue: https://github.com/wordpress-mobile/WordPress-Android/issues/68

The 'Default Image width' selector uses a bunch of predefined dimensions in it. I think it should show the predefined dimensions the user has set on the blog options.
Something like this:
- Original size
- Max (actual dimensions read from options)
- Medium (actual dimensions read from options)
- Small (actual dimensions read from options)
- Custom (Maybe?)

rachelmcr:
This issue specifically mentions the "Default Image Width" (in the blog settings) but can also apply to the image settings in the editor.

My Site > Settings > Default Image Width:

![2016-03-23 15 19 20](https://cloud.githubusercontent.com/assets/8658164/13990196/550aa300-f10b-11e5-897f-d743749dd690.png)

Editor > Image settings:

![2016-03-23 15 20 16](https://cloud.githubusercontent.com/assets/8658164/13990216/6cd7909c-f10b-11e5-843a-ef8b8d35fb4a.png)

I agree with the suggestion to use the image size settings from the blog's media settings (e.g. Thumbnail, Medium, Large), Original/Full size, and Custom (perhaps custom only in the editor's image settings, for resizing on the fly). 

For comparison, the iOS app uses a Maximum Image Upload Size setting in the Account Settings with a slider, and in the editor it offers a preset list (the same for all blogs, not based on the blog settings) of image sizes to choose from:

![2016-03-23 15 23 13](https://cloud.githubusercontent.com/assets/8658164/13990557/c5ae2ba8-f10c-11e5-8351-c76824d53f28.png)

Tested in Nexus 9, Android 6.0.1, alpha-4.

Issue #16, opened 2015-04-01
Featured image


aforcier:
Handle setting, modifying and removing the featured image.

Setting the featured image is currently part of `Image Settings` on Android, while on iOS it's contained in `Post Settings`.

Related issues:
[WordPress-Editor-Android#13](https://github.com/wordpress-mobile/WordPress-Editor-Android/issues/13)
[WordPress-Editor-Android#6](https://github.com/wordpress-mobile/WordPress-Editor-Android/issues/6)
https://github.com/wordpress-mobile/WordPress-Android/issues/169

aforcier:
Relevant (adds a reference to the featured image's id to the `posts` table of WPAndroid): https://github.com/wordpress-mobile/WordPress-Android/pull/2916

aforcier:
Ideally, we can allow each individual image to be set as featured through its own image settings, but also allow the featured image to be set/changed through `Post Settings` as on iOS.

aforcier:
This one's nearly ready for a PR, but I'd like some input before I open it.

First, here's the current featured image flow with the legacy Android editor:

https://cloudup.com/c4olwoeZVzU

The featured image is set by tapping on an image in the post and checking a box (only possible for local drafts on legacy).

This is different from iOS, where the featured image is set in the post settings instead:

https://cloudup.com/cM-r3kN-ofN

This is more similar to the web than the Android legacy implementation is, where you set the featured image from the sidebar:

![16-featured-image-web](https://cloud.githubusercontent.com/assets/9613966/11022688/d7e2e312-8632-11e5-8a15-7ae7bf4c07b6.png)

For our new editor, I'm trying out implementing both methods (image-specific settings and post settings). I personally prefer the image-specific setting flow (as on Android legacy) on mobile, and in any case it's probably good to keep functionality that users are used to from legacy where possible.

<h3>Setting the featured image from image settings</h3>

Here's the flow for changing the featured image through image-specific settings (PR forthcoming):

https://cloudup.com/caRCGPOBaxV

One change from legacy is that I dropped the second checkbox: 'Include image in post content'. In the legacy editor, if you leave that checkbox unchecked, the image stays in the draft, but upon uploading the post the featured image would be set and the image stripped from the contents of the post.

For the new editor I think we can drop that checkbox, for two reasons:
[...]

PR #244, opened 2015-09-09
Feature #13: Edit images

aforcier:
Addresses #13. Adds support for editing existing images to the editor.

Tapping an image shows an overlay:

13-edit-image-overlay

Tapping the overlay displays a modal dialog with image options:

13-edit-image-settings-pr

13-edit-image-settings-pr-2

<h3>Notes</h3>

<h4>1. Color scheme update</h4>

With this PR I also brought in the `colorControlActivated` color scheme from WPAndroid for the example app, to make them more similar visually. This changes the color of active EditTexts (see e.g. the title field in the image settings dialog above) and the cursor tab color from the default green to WP Light Blue:

editor-tab-color-update

<h4>2. No featured image support yet</h4>

For now, the two featured image checkboxes at the bottom don't do anything - featured image support will be added separately (#16)

<h4>3. Predef image sizes vs slider</h4>

Eventually we'll probably want to use predefined image sizes (Thumbnail, Medium, Large, Full, Custom?), for parity with wp-admin and the iOS editor. Since WPAndroid doesn't currently retrieve and store that information, I'm deferring it to be handled separately (pre-existing issue, imported from WPAndroid repo: #112).

Instead, I imported the legacy editor's image width slider:

13-edit-image-settings-slider

The maximum width (for all images uploaded to that WP site) is set by the user in the WPAndroid settings (in the example app, it's hardcoded to `600px` for now).

More details and screenshots here: https://github.com/wordpress-mobile/WordPress-Editor-Android/issues/13#issuecomment-138339101

<h4>4. Known issue: missing blur effect on API<19</h4>

The blurring effect for the edit image overlay doesn't work on API<19:

13-edit-images-overlay-blur-comparison

This is because the Android WebKit (which powers the WebView pre-API 19) doesn't support CSS filter effects (applied with `-webkit-filter`):

http://caniuse.com/#feat=css-filters

(this also happens to the failed upload overlay on those versions)


aforcier:
This PR also fixes an existing issue with image alignment on the `ZSSEditor` side. Image alignment was being reset when opening the editor (pre-existing iOS editor issue: https://github.com/wordpress-mobile/WordPress-Editor-iOS/issues/612).

It looks like the problem was that the editor pulls the alignment information from the caption when editing image settings, even if a caption isn't set. This works if a caption exists, since WordPress stores the image alignment data in the caption shortcode instead of inside the image element in that case.

In cases where there is no caption, the `ZSSEditor` finds the caption alignment information empty but ignores the alignment information in the image element. Fixed in [this commit](https://github.com/wordpress-mobile/WordPress-Editor-Android/commit/c250f38048ef7985ed569a45a7785a6a58b30b0c).

I haven't tested the fix on the Android app yet, but I was able to reproduce the bug in the example app by opening on image's options, changing the alignment, saving and re-opening it (the alignment information was wiped).

I'll add a comment about this to the iOS editor repo issue.

aforcier:
cc @bummytime, @thianhlu

thianhlu:
:+1:

bummytime:
@aforcier I noticed an issue on API 21. Looks like we lose all of the image (and post) data after the HTML button is tapped. Could be an existing issue that I just noticed now....

Video: https://cloudup.com/c3AbT7cFAM9

bummytime:
Tested on:
- API 16 :+1: 
- API 18 :+1: 
- API 19 :+1: 
- API 21 :-1: (Issue noted above)

Unit tests: Green
Connected Tests: Green

bummytime:
@thianhlu Regarding:

> 1. Known issue: missing blur effect on API<19
> 
> The blurring effect for the edit image overlay doesn't work on API<19:

that @aforcier mentioned above, is this a showstopper for you? I know part of the reason we were blurring and changing the opacity was to help make the "Edit"  icon more prominent...especially with "whiter" bg images.

thianhlu:
Can we do a dark semi transparent overlay on <19 instead?

aforcier:
> I noticed an issue on API 21. Looks like we lose all of the image (and post) data after the HTML button is tapped. Could be an existing issue that I just noticed now....

I think I've seen this happen before on the API 21 emulator - `getContent()` or `getTitle()` are taking too long to return the HTML from the visual editor. Within WPAndroid, this would fall back to the last auto-save (every 10 seconds), but in the example app we don't auto-save so it falls back to the last saved title and content native-side (in your case, they're empty).

This is _probably_ caused by the slow and laggy WebView in the API 21 emulators, and might not occur on real devices.

But we could rethink what happens if for some reason `getContent()` or `getTitle()` take longer than 1 second each (the arbitrary time-out I set for them). It should take much less on most devices, so we could probably bump that up to 5 seconds and abort/show a toast if it still doesn't return. Since we haven't had this happen on a real device it's hard to know how extensive the issue is.

I can open an issue for looking into this. We can also try to troubleshoot the API 21 emulator, in case we can learn something from that.

aforcier:
> Can we do a dark semi transparent overlay on <19 instead?

Got some CSS advice from @v18 and looked into some options for older APIs. How is this @thianhlu (API 18)?

13-edit-image-api18-dark-overlay

(`0.5` opacity black overlay)

bummytime:
@aforcier 

> This is probably caused by the slow and laggy WebView in the API 21 emulators, and might not occur on real devices.

Makes sense to me. Now that you say this I recall us talking about the API 21 emu giving us fits..getting old...poor memory. :older_man: 

> I can open an issue for looking into this. We can also try to troubleshoot the API 21 emulator, in case we can learn something from that.

I am good with opening a new issue.

bummytime:
> Got some CSS advice from @v18 and looked into some options for older APIs. 

Yay @v18 ... I'm glad somebody here knows CSS well ;-)

enhanced-14836-1414320930-8

libs/editor-common/assets/ZSSRichTextEditor.js
```diff
@@ -1077,7 +1077,8 @@ ZSSEditor.updateCurrentImageMeta = function( imageMetaString ) {
     // elements surround the targeted node.  This approach is safer.
     var node = ZSSEditor.findImageCaptionNode( ZSSEditor.currentEditingImage );
     node.insertAdjacentHTML( 'afterend', html );
-    node.remove();
+    // Use {node}.{parent}.removeChild() instead of {node}.remove(), since Android API<19 doesn't support Node.remove()
+    node.parentNode.removeChild(node);
```
bummytime, line 1081: Nice find! comment++

WordPressEditor/src/main/java/org/wordpress/android/editor/ImageSettingsDialogFragment.java
```diff
@@ -0,0 +1,261 @@
...
+            return Math.min(imageWidthPictureSetting, imageWidthBlogSetting);
+        }
+    }
+
+    /**
+     * Return the integer value of the width EditText, adjusted to be within the given min and max, and stripped of the
+     * 'px' units
+     */
+    private int getEditTextIntegerClamped(EditText editText, int minWidth, int maxWidth) {
+        int width = 10;
```
bummytime, line 328: Magic number?

bummytime:
@aforcier Ok. code looks good overall. 

I have one more general comment about the UX... and you or @thianhlu can tell me to take a long walk off a short pier if you want since you prolly had many discussions about this.

I think the dialog seems a little unwieldy — lots of options and different controls in a narrow space where it may not be obvious the content in the dialog can scroll. Google mentions this in their [design guidelines](https://www.google.com/design/spec/components/dialogs.html#dialogs-content):

> Avoid dialogs that:
> 
> Contain scrolling content, particularly alerts. Instead, consider alternate containers or layouts that are optimized for reading or interacting with significant amounts of content.

I am wondering if it would be a better approach to use a full screen dialog here:

scaled components_dialogs_fullscreen1

We seem to meet most of the criteria that [Google provides](https://www.google.com/design/spec/components/dialogs.html#dialogs-full-screen-dialogs):

> Consider using a full-screen dialog when the content or process meets any of the following criteria:
> - The dialog includes components like pickers or form fields requiring an input method editor (IME), such as a keyboard.
> - When changes are not saved in real time
> - When there is no draft capability in the app
> - When performing batch operations or queuing changes prior to submitting them

and it would provide additional space vertically and horizontally for layout. This is one of the reasons we did a full screen modal in iOS.

Again, take it with a grain of salt...just my 2 pesos.

thianhlu:
awesome guys! looks good

thianhlu:
On both the overlay and full screen dialog

WordPressEditor/src/main/java/org/wordpress/android/editor/ImageSettingsDialogFragment.java
aforcier, line 328: Good point, I'll switch that to a constant.

aforcier:
Here's how the dialog is looking in fullscreen:

13-edit-images-fullscreen

Not all of the plumbing's hooked up yet, but here are the changes I made so far:
- Added close and save buttons to the action bar
- Changed the background from pure white to WP Gray Light, to match the other fullscreen dialogs in WPAndroid
- Used `16dp` instead of `24dp` side padding for the dialog, and added the same padding on the top and bottom

Re: the padding, [Google suggests](https://www.google.com/design/spec/layout/metrics-keylines.html#metrics-keylines-keylines-spacing) `16dp` on mobile for the sides. This isn't specific to dialogs though, and we can revert/modify this. The specs specify `24dp` in the case of tablets, though we may want much more than that in that case (see below).

I'm also experimenting with adding a thumbnail of the image at the top along with its filename, like we're doing on iOS. With a fullscreen dialog you can't see where you are in the document in the background anymore, so I think having some context for which image's settings you're editing is extra important now.

13-edit-images-fullscreen-with-thumbnail

One other thing worth considering is how modal vs fullscreen looks on tablets:

13-edit-images-tablet-comparison

We may also want to add some extra padding on the sides on tablets if we're keeping fullscreen on those devices.

Opinions on using the thumbnail/layout nits/tablet thoughts @thianhlu @bummytime?

thianhlu:
@aforcier After our discussion today, I think we can make full screen dialog work. Can we try a few more tweaks. Apologies for oversimplified mock but its mainly a spacing issue right now. Left and right spacing is the same. Other change is the background is white now.

aforcier:
Added the 'Do Not Merge' tag until this is ready for another review. Will be pushing commits as I work on the following:
- Load a thumbnail in image settings dialog
- Tablet version of the image settings dialog, with special padding rules
- Semi-transparent dark overlay for selected images on `API < 19`

aforcier:
Ready for another look, @thianhlu & @bummytime!

<h3>Summary of latest changes</h3>
- Added thumbnail support
- Added special rules for using a dark semi-transparent overlay when CSS filter effects (which include blur) are not supported
- Customized the image settings dialog for tablets (width `720dp`+), with the spacing changes outlined above
- Fixed a few ActionBar-related bugs the full-screen dialog introduced
- Sanity tested the fragment work on WPAndroid, no big issues (some small ones which will be addressed in the integration PR)

<h3>Dialog design changes</h3>
- Used `104dp` padding, with 'Image Settings' title aligned with dialog contents
- Reverted back to white background

Phone (back to white background):

13-edit-images-phone-white-bg

Tablet:

13-edit-images-tablet-v2

The spacing on the right of the Save button is still off. The best way to fix this would be to use a drawable instead of text for that button.

BTW, this also happens to the Publish/Update button on WPAndroid, so we may prefer to fix both or neither for consistency:

wpandroid-edit-post-tablet

thianhlu:
@aforcier Yeah lets fix spacing on Save btn. Also, lets update the navBar spacing on editor as well (left and right sides) on tablet view. Otherwise all else looks great. Thanks for getting these tweaks in!

aforcier:
Switched the ActionBar to use a custom layout - now there's some extra padding:

13-edit-images-tablet-save-button-custom-view

The spacing isn't perfect yet, but from after discussion with @thianhlu we decided to open a separate issue for fixing that (and another for porting these changes to the editor proper).

thianhlu:
:+1:

aforcier:
Opened some issues for adjusting the right padding for the ActionBar in the image settings dialog (#252), porting these changes to the editor fragment ActionBar proper (#253), and investigating the `API21` emulator weirdness (#254).

bummytime:
Tested on:
- API 16 :+1: 
- API 18 :+1: 
- API 19 :+1: 
- API 21 :+1:

Unit tests: Green
Connected Tests: Green

Looking good @aforcier!

Issue #247, opened 2015-09-10
Add a button for debugging HTML


aforcier:
Since remote debugging (enabled using `WebView.setWebContentsDebuggingEnabled()`) wasn't added until `API 19`, we're a little bit in the dark when debugging HTML/JS weirdness on older APIs (which tend to have most of the bugs).

We should add a debug-mode only button which outputs the raw HTML to the console, to help with debugging on those early APIs.

PR #248, opened 2015-09-10
Feature #247: Log HTML button

aforcier:
Implements #247.
- Adds a debug flag to the editor, which has to be enabled by the host activity to turn on debug features (turned off by default)
- When debug mode is active, adds a button to the ActionBar which prints the raw HTML from the editor into the console (without performing the usual formatting through `wp.saveText`)

247-html-debug


aforcier:
cc @bummytime

bummytime:
:shipit:

PR #266, opened 2015-11-24
Import featured image support from WPAndroid

aforcier:
Imports the changes made on the WPAndroid repo for featured images: https://github.com/wordpress-mobile/WordPress-Android/pull/3425

Also fixes an error in the example activity, and repairs Travis to work with the new build tools and gradle version as updated from WPAndroid.

cc @bummytime

PR #267, opened 2015-11-24
Issue #261: Image settings screen navigation fixes

### LINKED ISSUE #261: Image settings screen navigation fixes
Two navigation issues:
1. When tapping the device back button, the app exits the editor instead of just the image settings screen - this is a bug, we should be returning to the editor
2. From conversation with @maxme, we may be risking users pressing X and expecting their changes to be saved in the image settings screen:

image setttings

Even though the X is unique, in the rest of the app settings changes are persisted when the user hits the back arrow and exits the settings screen, so the user may press the X by habit.

The fix would be to drop the 'SAVE' button entirely and switch the X back to a back arrow, which saves any changes and returns to the editor.

The Android back arrow for reference:

wpandroid-editor-0 4-screen

cc @thianhlu @bummytime @meremagee @maxme

aforcier:
Fixes #261, making a few improvements to navigating away from the image settings dialog.
- Tapping the device back button now returns the user to the editor (instead of exiting the editor entirely), making it the same as pressing the home button in the ActionBar
- Show a confirmation dialog when closing the settings dialog with unsaved changes:

261-image-settings-nav-dialog
- Show a toast that changes have been saved when the Save button is pressed in the ActionBar:

261-image-settings-nav-saved-toast
- Also incorporates checking for changes to featured images, which were introduced in #266

Demo of all the changes, on WPAndroid: https://cloudup.com/cYBMtWClmuz

Testing note: The image settings dialog should catch any changes to the featured image checkbox and prompt you with the dialog. However, since the demo app doesn't handle featured images right now, the change isn't reflected in the editor itself when the screen is closed (so all images are marked as not featured when their settings dialog is opened).

This will require a small change to WPAndroid when milestone `0.5` is merged in (the same one made to `EditorExampleActivity` [here](https://github.com/wordpress-mobile/WordPress-Editor-Android/commit/396ffcdb124e11a9b741d43e6184c37e8781334a#diff-0be19c8bcb324d950b8ac192edc6a0adR65). After this PR is merged, let's keep the #261 issue open to track that (but label it done).

cc @bummytime, @thianhlu, @meremagee

### DIFF
--- a/WordPressEditor/src/main/java/org/wordpress/android/editor/ImageSettingsDialogFragment.java
+++ b/WordPressEditor/src/main/java/org/wordpress/android/editor/ImageSettingsDialogFragment.java
@@ -96,8 +96,12 @@ public class ImageSettingsDialogFragment extends DialogFragment {
             public void onClick(View v) {
                 mImageMeta = extractMetaDataFromFields(mImageMeta);
 
-
-                imageRemoteId = mImageMeta.getString("attachment_id");
+                String imageRemoteId = "";
+                try {
+                    imageRemoteId = mImageMeta.getString("attachment_id");
+                } catch (JSONException e) {
+                    AppLog.e(AppLog.T.EDITOR, "Unable to retrieve featured image id from meta data");
+                }
 
                 Intent intent = new Intent();
                 intent.putExtra("imageMeta", mImageMeta.toString());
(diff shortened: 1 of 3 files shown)


WordPressEditor/src/main/java/org/wordpress/android/editor/ImageSettingsDialogFragment.java
```diff
@@ -229,6 +269,50 @@ private void restorePreviousActionBar() {
...
+        });
+
+        AlertDialog dialog = builder.create();
+        dialog.show();
+    }
+
+    /**
+     * Extracts the meta data from the dialog fields and updates the entries in the given JSONObject.
+     */
+    private JSONObject extractMetaDataFromFields(JSONObject metaData) {
```
bummytime, line 298: :+1:

PR #270, opened 2015-11-29
Issue #242: Address lower API CSS issues with uploading image containers

### LINKED ISSUE #242: Create separate CSS doc to address issues in lower API levels
Title has it. Problems with current CSS doc:

<h4>Missing upload progress bar (API < 19)</h4>

On all `API < 19` devices, the progress bar doesn't show up when uploading images. It seems that Android didn't support the `<progress>` tag we're using until the Chromium WebView (`API 19`).  One possibility here is to use an upload notification instead. The WPAndroid app uses those for post uploads and gallery uploads, so we might want to use it for individual image uploads anyway (on all APIs, even ones where the progress bar inside the WebView is shown). Here's a demo of what that would look like on an API 18 device (the notification isn't included in this PR, just looking for feedback):

https://cloudup.com/cqzRAsg_CeE

<h4>Uploading image container is sometimes very small (API < 19)</h4>

Sometimes (probably when uploading media in a blank document, still testing) the image container is very small while uploading. Once the container is removed at the end of the upload, the normal-sized image is shown:

https://cloudup.com/c2HhsInzXsW

The problem is resolved by removing [this](https://github.com/wordpress-mobile/WordPress-Editor-Android/blob/244a51315e24eddc0510b422a75e6483c6c0f6f3/libs/editor-common/assets/editor.css#L141) line from the CSS. However, this probably isn't the root cause, and it also breaks the progress bar on `API > 18`.

Ref: #236

aforcier:
Fixes #242, addressing the following bugs that appear on `API < 19` only:
- Fixes the issue where the uploading image container would sometimes be very small: https://cloudup.com/c2HhsInzXsW
- Because the `API < 19` `WebView` doesn't support the progress tag, there is no indication that the image is uploading. For those API levels, an 'Uploading...' overlay is shown instead:

242-upload-overlay-text-old-api

Some explanation about 9e1863f: Fixing the initial issue with the small containers required using `display: block` instead of `display: inline-block`. But this makes the container fill the screen width rather than match the image size, so for images less wide than the screen the overlays are skewed:

242-uploading-image-block-bug-api18

I found that once the image has loaded (on progress updates after the initial one which merely initializes the progress bar), we can safely drop back to `inline-block` on those API levels. This keeps the small container fix working, but also properly centers the overlays on the image for smaller images:

242-uploading-image-inline-block-api18

It's not an ideal fix as we may occasionally see a brief alignment issue until the first progress update comes in, but I haven't found a better way to fix both things at once so far.

cc @bummytime @thianhlu

### DIFF
--- a/libs/editor-common/assets/ZSSRichTextEditor.js
+++ b/libs/editor-common/assets/ZSSRichTextEditor.js
@@ -1028,6 +1028,9 @@ ZSSEditor.markImageUploadFailed = function(imageNodeIdentifier, message) {
     if (imageProgressNode.length != 0){
         imageProgressNode.addClass('failed');
     }
+
+    // Delete the compatibility overlay if present
+    imageContainerNode.find("span.upload-overlay").addClass("failed");
 };
 
 /**
@@ -1051,6 +1054,9 @@ ZSSEditor.unmarkImageUploadFailed = function(imageNodeIdentifier, message) {
     if (imageProgressNode.length != 0){
         imageProgressNode.removeClass('failed');
     }
+
+    // Display the compatibility overlay again if present
+    imageContainerNode.find("span.upload-overlay").removeClass("failed");
 };
 
 /**
--- a/libs/editor-common/assets/ZSSRichTextEditor.js
+++ b/libs/editor-common/assets/ZSSRichTextEditor.js
@@ -946,6 +946,12 @@ ZSSEditor.setProgressOnImage = function(imageNodeIdentifier, progress) {
         imageNode.addClass("uploading");
     }
 
+    // Revert to non-compatibility image container once image upload has begun. This centers the overlays on the image
+    // (instead of the screen), while still circumventing the small container bug the compat class was added to fix
+    if (progress > 0) {
+        this.getImageContainerNodeWithIdentifier(imageNodeIdentifier).removeClass("compat");
+    }
+
     var imageProgressNode = this.getImageProgressNodeWithIdentifier(imageNodeIdentifier);
     if (imageProgressNode.length == 0){
           return;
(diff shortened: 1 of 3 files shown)


libs/editor-common/assets/ZSSRichTextEditor.js
```diff
@@ -932,6 +946,12 @@ ZSSEditor.setProgressOnImage = function(imageNodeIdentifier, progress) {
         imageNode.addClass("uploading");
     }
 
+    // Revert to non-compatibility image container once image upload has begun. This centers the overlays on the image
+    // (instead of the screen), while still circumventing the small container bug the compat class was added to fix
+    if (progress > 0) {
+        this.getImageContainerNodeWithIdentifier(imageNodeIdentifier).removeClass("compat");
```
bummytime, line 961: :sparkles: :ghost:

bummytime:
Nice fix @aforcier. It runs well on a variety of low-API devices and code looks fine. The only comment I have is regarding the text on a light image:

scaled screen shot 2015-11-30 at 3 06 23 pm

Can be difficult to see. Any thoughts?

aforcier:
> The only comment I have is regarding the text on a light image...can be difficult to see. Any thoughts?

I think we can use a semi-transparent dark overlay over the image while uploading again, to increase the contrast for the `Uploading...` message. This is what we did for the [edit overlay](https://github.com/wordpress-mobile/WordPress-Editor-Android/pull/244#issuecomment-139531939).

13-edit-image-api18-dark-overlay

I have this mostly working, there are some CSS quirks to resolve first. I'll update this branch when I get that working.

aforcier:
Added a dark semi-transparent overlay on `API<19` for uploading and failed images.

Ready for another look @bummytime!

tonyr59h:
Looks good, @aforcier! Working as expected above and below the API 19 line.

:shipit:

PR #278, opened 2016-01-28
Fix #183: Visual and Legacy editors accessibility

### LINKED ISSUE #183: Verify accessibility
Placeholder issue - test the app for accessibility.

maxme:
Fix #183: Visual and Legacy editors accessibility
- [x] Label the format bar buttons
- [x] Change the title of the HTML file to something more appropriate ("Visual Editor" perhaps)
- [x] The post and image settings screens both need accessibility labels for the various fields (small changes here, image settings was pretty easy to use with Talkback)
- [x] <s> Related to the above, we should make the switch between visual/HTML mode more obvious. Maybe a toast ("Switched to visual mode", "Switched to HTML mode"), those seem to be read out automatically (this also came up in the iOS editor issue above).</s>  (Not implemented)

I've tried to avoid `%%TITLE%%` text replacement, but none actually worked:
- `mWebView.execJavaScriptFromString("document.title = 'Test';");`
- `mWebView.setContentDescription("Test");` 
  (same results with or without the `<title>` in the html file)

Alternative solution is to remove the title completely, not sure this is a problem for our users.

Related https://github.com/wordpress-mobile/WordPress-Android/pull/3666

### DIFF
--- a/WordPressEditor/src/main/res/values/strings.xml
+++ b/WordPressEditor/src/main/res/values/strings.xml
@@ -65,7 +65,7 @@
     <string name="format_bar_description_bold">Bold</string>
     <string name="format_bar_description_italic">Italic</string>
     <string name="format_bar_description_underline">Underline</string>
-    <string name="format_bar_description_strike">Strike</string>
+    <string name="format_bar_description_strike">Strikethrough</string>
     <string name="format_bar_description_quote">Block quote</string>
     <string name="format_bar_description_link">Insert link</string>
     <string name="format_bar_description_more">Insert more</string>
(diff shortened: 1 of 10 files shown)


WordPressEditor/src/main/res/values/strings.xml
```diff
@@ -61,4 +61,19 @@
         <item>Full Size</item>
     </string-array>
 
+    <!-- Accessibility - format bar button descriptions -->
+    <string name="format_bar_description_bold">Bold</string>
+    <string name="format_bar_description_italic">Italic</string>
+    <string name="format_bar_description_underline">Underline</string>
+    <string name="format_bar_description_strike">Strike</string>
```
aforcier: Maybe 'Strikethrough' would be clearer than 'Strike' here?

aforcier:
As discussed, we'll open a separate issue for the remaining issues in #183 (mainly, notifying users about switching to legacy if the visual editor isn't able to meet their accessibility needs).

:shipit:

PR #280, opened 2016-02-05
Remove analytics - use an optional listener instead

maxme:
We should discuss this - not sure we should merge.

(Optional listener to use in wpandroid: https://gist.github.com/maxme/8600cdf07f63f33f5490)


maxme:
Updated with `onTrackableEvent`

maxme:
wpandroid callback: https://gist.github.com/maxme/5b366a79dec750ed1ca7

WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragment.java
```diff
@@ -740,13 +738,13 @@ public void run() {
                 if (URLUtil.isNetworkUrl(mediaUrl)) {
                     String mediaId = mediaFile.getMediaId();
                     mWebView.execJavaScriptFromString("ZSSEditor.insertImage('" + mediaUrl + "', '" + mediaId + "');");
-                    AnalyticsTracker.track(Stat.EDITOR_ADDED_PHOTO_VIA_WP_MEDIA_LIBRARY);
+                    mEditorFragmentListener.onTrackableEvent(TrackableEvent.NETWORK_MEDIA_ADDED);
```
aforcier: We should use `NETWORK_IMAGE_ADDED` and `LOCAL_IMAGE_ADDED` instead of `_MEDIA_`, since we're going to need to differentiate between images and video when we add video insert/upload support.

maxme: > We should use NETWORK_IMAGE_ADDED and LOCAL_IMAGE_ADDED instead of _MEDIA_, since we're going to need to differentiate between images and video when we add video insert/upload support.

Right, that's weird I feel like I already did that change in WPAndroid

maxme: Ok, I can't read correctly... `(AnalyticsTracker.track(Stat.EDITOR_ADDED_PHOTO_VIA_WP_MEDIA_LIBRARY);)` was done ;)

aforcier: I mean that `TrackableEvent` should have `NETWORK_IMAGE_ADDED` instead of `NETWORK_MEDIA_ADDED`, and the same for `LOCAL_`. It's already mapping to the correct `Stat` in the gist:

https://gist.github.com/maxme/5b366a79dec750ed1ca7#file-ontrackableevent-java-L13-L18

It should just be more specific in the editor's `TrackableEvent` so we can add in separate video events later, which will map to the already existing `EDITOR_ADDED_VIDEO_VIA_WP_MEDIA_LIBRARY` and `EDITOR_ADDED_VIDEO_VIA_LOCAL_LIBRARY`.

maxme: Renamed Media/Image events in 69aaa8a

As per our chat, I'll probably kill `NETWORK_IMAGE_ADDED`, `LOCAL_IMAGE_ADDED`, `UPLOAD_IMAGE_RETRIED` and `UPLOAD_IMAGE_FAILED` and track them in the activity.

maxme: Note: events killed while working on - https://github.com/wordpress-mobile/WordPress-Android/issues/3610 -

aforcier:
:shipit:

PR #283, opened 2016-02-11
fix #282: localize JS strings by using nativeState js-java interface

### LINKED ISSUE #282: Localize JS strings
`Edit` and `Tap to try again` are hardcoded strings in the `ZSSEditor` and cannot be localized. We'll probably need to assign those strings on init, so we can pass a localized String from the native side.

Counterpart issue in the iOS editor repo: https://github.com/wordpress-mobile/WordPress-Editor-iOS/issues/756

maxme:
fix #282: localize JS strings by using nativeState js-java interface

### DIFF
--- a/WordPressEditor/src/main/res/values/strings.xml
+++ b/WordPressEditor/src/main/res/values/strings.xml
@@ -12,7 +12,7 @@
     <string name="edit">Edit</string>
     <string name="tap_to_try_again">Tap to try again!</string>
     <string name="uploading">Uploading…</string>
-    <string name="uploading_gallery_placeholder">[Uploading Gallery]</string>
+    <string name="uploading_gallery_placeholder">Uploading gallery…</string>
 
     <string name="alert_insert_image_html_mode">Can\'t insert media directly in HTML mode. Please switch back to visual mode.</string>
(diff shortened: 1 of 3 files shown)


WordPressEditor/src/main/res/values/strings.xml
```diff
@@ -9,6 +9,10 @@
     <string name="delete">Delete</string>
     <string name="save">Save</string>
     <string name="discard">Discard</string>
+    <string name="edit">Edit</string>
+    <string name="tap_to_try_again">Tap to try again!</string>
+    <string name="uploading">Uploading…</string>
+    <string name="uploading_gallery_placeholder">[Uploading Gallery]</string>
```
aforcier: Could we keep 'Uploading gallery...' as the message? Also it's probably better to remove the square brackets from the string and add wrap them around the method call in JavaScript instead - that way we can reuse this string in the future if we add visual gallery support and can drop the shortcode in visual mode.

aforcier:
Looks good @maxme, good to :shipit: once that line note is addressed.

maxme:
Fixed in 166cf1e

aforcier:
:shipit:

PR #289, opened 2016-02-23
Issue #24: addJavascriptInterface security fix

### LINKED ISSUE #24: addJavascriptInterface security check
Verify we're not vulnerable to `WebView.addJavascriptInterface` [reflection exploits](http://developer.android.com/reference/android/webkit/WebView.html#addJavascriptInterface%28java.lang.Object,%20java.lang.String%29) on `API<17`.
- [x] Check loading post data from .org sites with raw JavaScript in the content
- [x] Check previewing (should not render any JS in the post being previewed)

aforcier:
Fixes #24.
- On `API<17`, replaced the `nativeCallbackHandler` `JavascriptInterface` with running JS callbacks through a sandboxed iframe, which are received by `WebViewClient.shouldOverrideUrlLoading` and piped through to the `JsCallbackReceiver`
- Dropped `NativeStateJsInterface` for all APIs. Instead, the native API level and localized Strings are injected when loading the `android-editor.html` file
- Added profiling markers to track editor startup time

### DIFF
--- a/WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragment.java
+++ b/WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragment.java
@@ -377,7 +377,15 @@ public class EditorFragment extends EditorFragmentAbstract implements View.OnCli
             htmlEditor = htmlEditor.replace("%%TITLE%%", getString(R.string.visual_editor));
         }
 
-        mWebView.addJavascriptInterface(new JsCallbackReceiver(this), JS_CALLBACK_HANDLER);
+        // To avoid reflection security issues with JavascriptInterface on API<17, we use an iframe to make URL requests
+        // for callbacks from JS instead. These are received by WebViewClient.shouldOverrideUrlLoading() and then
+        // passed on to the JsCallbackReceiver
+        if (Build.VERSION.SDK_INT < 17) {
+            mWebView.setJsCallbackReceiver(new JsCallbackReceiver(this));
+        } else {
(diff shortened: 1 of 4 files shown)


WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragment.java
```diff
@@ -372,14 +374,26 @@ protected void initJsEditor() {
...
+                    "nativeState.localizedStringUploadingGallery = '" + getString(R.string.uploading_gallery_placeholder) + "';\n");
         }
 
-        mWebView.addJavascriptInterface(new JsCallbackReceiver(this), JS_CALLBACK_HANDLER);
-        mWebView.addJavascriptInterface(new NativeStateJsInterface(getActivity().getApplicationContext()),
-                                        JS_STATE_INTERFACE);
+        // To avoid reflection security issues with JavascriptInterface on API<17, we use an iframe to make URL requests
+        // for callbacks from JS instead. These are received by WebViewClient.shouldOverrideUrlLoading() and then
+        // passed on to the JsCallbackReceiver
+        if (Build.VERSION.SDK_INT < 17) {
```
maxme, line 392: I hope custom ROMs are not messing with these version codes, but in case they do, we should maybe use `Build.VERSION_CODES.JELLY_BEAN_MR1` instead of `17`.

maxme, line 392: scratch that - we can't reference `Build.VERSION_CODES` on the JS side

maxme:
:shipit:

PR #304, opened 2016-02-29
fix #281: drop OpenSans use default sans serif system font (Roboto) instead

### LINKED ISSUE #281: Drop Open Sans
To reflect upcoming changes to the web version, we should drop Open Sans and use system fonts (Roboto) instead.

maxme:
fix #281: drop OpenSans use default sans serif system font (Roboto) instead


aforcier:
Tested on APIs `15`, `18`, `23` to sanity-check, looks good. :shipit:

PR #306, opened 2016-03-01
Issue/268 link dialog field text

### LINKED ISSUE #268: Improve link pasting experience in link dialog
Original issue: https://github.com/wordpress-mobile/WordPress-Android/issues/3443

> When I make a bunch of text into a hyperlink in the Android app, the text input field for the link is pre-loaded with `http://` which is actually kind of inconvenient, since almost all links on mobile are copy-pasted from my mobile browser window. Since the field doesn't filter out the double `http://` I have to delete it every time.

The link dialog in the new editor has the same issue as legacy. We should duplicate the iOS app behavior, which auto-fills the URL field with the URL in the clipboard. At the very least, we should filter out double `http://` when pasting in a URL.

tonyr59h:
Fix #268 by removing the default `http://` string from the URL field and filling in any existing clipboard URL.

Additionally 4e51cf8c413e248702ada791d2a4d11530422417 changes the selection of the URL field from the end of the text to selecting all text. This seems more useful (user can quickly delete if desired) and better aligns the Android functionality with iOS (iOS field has a delete all button).

I noticed the keyboard was not showing if it was hidden when pressing the Link button. d6d5868c123ff4c3004865015a8b21bba76325a4 will ensure that the keyboard is always shown when the dialog is displayed.

One more question before merging is in the issue, copied here for convenience:

> After toying around I think it would be nice if the link text field would auto-fill with whatever text your cursor is in the middle of. Currently it will only fill that field if text is highlighted. Thoughts?

### DIFF
--- a/WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragment.java
+++ b/WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragment.java
@@ -7,6 +7,7 @@ import android.content.Context;
 import android.content.DialogInterface;
 import android.content.Intent;
 import android.content.res.Configuration;
+import android.net.Uri;
 import android.os.Build;
 import android.os.Bundle;
 import android.os.Looper;
@@ -16,6 +17,7 @@ import android.support.v7.app.AppCompatActivity;
 import android.text.Editable;
 import android.text.SpannableString;
 import android.text.Spanned;
+import android.text.TextUtils;
 import android.view.LayoutInflater;
 import android.view.Menu;
 import android.view.MenuInflater;
@@ -573,6 +575,8 @@ public class EditorFragment extends EditorFragmentAbstract implements View.OnCli
             if (linkText == null || linkText.equals("")) {
                 linkText = linkUrl;
             }
+
+            if (TextUtils.isEmpty(Uri.parse(linkUrl).getScheme())) linkUrl = "http://" + linkUrl;
 
             if (mSourceView.getVisibility() == View.VISIBLE) {
                 Editable content = mSourceViewContent.getText();
--- a/WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragment.java
+++ b/WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragment.java
@@ -513,7 +513,7 @@ public class EditorFragment extends EditorFragmentAbstract implements View.OnCli
             }
 
             linkDialogFragment.setArguments(dialogBundle);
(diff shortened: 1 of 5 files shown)


maxme:
There is a problem when adding a url manually without a prefix: https://cloudup.com/crGEVVRSy2t 

Note, I'm note sure why the selected text was modified (apparently the problem was here before that patch)

aforcier:
Looks like the selected text is being ignored on `API<22`, we've only tested emulators so far though. This was definitely pre-existing.

aforcier:
> > One more question before merging is in the issue, copied here for convenience:
> 
> After toying around I think it would be nice if the link text field would auto-fill with whatever text your cursor is in the middle of. Currently it will only fill that field if text is highlighted. Thoughts?

Do you mean if I've typed 'hello world' and put my cursor somewhere inside 'world' (with nothing highlighted) it should pre-fill 'world'? I think that could be a nice addition, as long as it doesn't happen if you're touching whitespace on either side. I haven't looked at the link code in a while so I'm not sure if this might be overly complicated to add - I'd be fine with leaving it out of this PR if it's too involved.

tonyr59h:
> Do you mean if I've typed 'hello world' and put my cursor somewhere inside 'world' (with nothing highlighted) it should pre-fill 'world'?

Yup, exactly. It looks like that will have to be done JS side in `getSelectedText` as well as in the [EditorFragment](https://github.com/wordpress-mobile/WordPress-Editor-Android/blob/develop/WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragment.java#L491). Probably best if we open a separate issue.

tonyr59h:
Created #307 to track the selected word improvements.

tonyr59h:
> There is a problem when adding a url manually without a prefix

Been looking into this problem. If there's no URI specified (`http://` for example) the baseURI attribute of nodes inserted will be the same as the document (`file:///android_asset/` in this case).

That property is read-only so we can either scrub the URL when querying it or ensure that inserted nodes have a defined `http://` URI. Probably best to do the latter so the post data doesn't have any Android specific references.

tonyr59h:
> There is a problem when adding a url manually without a prefix

56b5dfa13ef58cb4b7432c88d8b0b76a63e30e0e should address that issue now.

aforcier:
I opened a separate issue for the selected text bug @maxme mentioned (wasn't introduced by this PR): #309.

aforcier:
Nice work @tonyr59h, good catch on the keyboard bug/fix. :shipit:

PR #311, opened 2016-03-04
Fix #310: Wrap DialogLink's EditText inside TextInputLayout

### LINKED ISSUE #310: Add text labels on top of EditTexts in the Link pop over
Because it's hard to distinguish which EditText is used for URL or text - happen specifically with pre-filled links or when opening existing posts.

screenshot-2016-03-04_11 31 44 163

maxme:
Fix #310: Wrap DialogLink's EditText inside TextInputLayout

https://cloudup.com/cCeK8uxWUaGa


aforcier:
:shipit:

Issue #316, opened 2016-03-09
Test visual editor with Android N multitasking


aforcier:

bummytime:
:fearful:

maxme:
App works pretty well in multi tasking mode. I got a crash on the editor when resizing the window (doesn't happen everytime I resize it), looks like a missing `isAdded()`:

```
AndroidRuntime: Process: org.wordpress.android, PID: 24903
AndroidRuntime: java.lang.IllegalStateException: Fragment EditorFragment{5986b2b} not attached to Activity
AndroidRuntime:     at android.app.Fragment.getResources(Fragment.java:811)
AndroidRuntime:     at android.app.Fragment.getString(Fragment.java:833)
AndroidRuntime:     at org.wordpress.android.editor.EditorFragment$12.run(EditorFragment.java:944)
AndroidRuntime:     at android.os.Handler.handleCallback(Handler.java:739)
AndroidRuntime:     at android.os.Handler.dispatchMessage(Handler.java:95)
AndroidRuntime:     at android.os.Looper.loop(Looper.java:148)
AndroidRuntime:     at android.app.ActivityThread.main(ActivityThread.java:5849)
AndroidRuntime:     at java.lang.reflect.Method.invoke(Native Method)
AndroidRuntime:     at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:763)
```

Something annoying though: when you edit your post in HTML mode and resize the window, EditActivity is restarted and you're switched back to the Visual mode.

maxme:
Sometimes after a window resize the editor lose all content and title.

PR #325, opened 2016-03-23
Fix a rare crash

maxme:
This crash is easy to reproduce on Android N, by resizing the window several times. Ref. https://github.com/wordpress-mobile/WordPress-Editor-Android/issues/316#issuecomment-198454879


aforcier:
Fixes the crash, :shipit:.

PR #326, opened 2016-03-23
Fix #314: Cursor jumping when switching from HTML to visual

### LINKED ISSUE #314: Cursor jumps to top right when switching from HTML to Visual mode
When switching between HTML and Visual modes, I expect the cursor to either stay where it is in the post or jump to the very beginning of the post.

Instead, if the cursor isn't on the first line of text when you switch from visual to html mode, it will appear in the top right corner when you switch back to visual mode.

**Steps to reproduce:**
1. In the visual editor, type some text, hit return to start a new line, and type more text.
2. With the cursor on the second line of text, switch to the html mode.
3. Switch back to the visual mode.

Result: The cursor appears in the top right corner of the editor. If you start typing without repositioning the cursor, it will place the text at the beginning of the line you were on before leaving the visual editor (even if that wasn't where the cursor was placed before switching between visual and html).

**Screenshot** (after switching from visual to html and back to visual):

Tested on Nexus 9, Android 6.0.1, app version alpha-2.

aforcier:
Fixes #314 by resetting the selection when returning from HTML mode to visual.

cc @maxme

### DIFF
--- a/WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragment.java
+++ b/WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragment.java
@@ -453,7 +453,8 @@ public class EditorFragment extends EditorFragmentAbstract implements View.OnCli
                 // Update the list of failed media uploads
                 mWebView.execJavaScriptFromString("ZSSEditor.getFailedMedia();");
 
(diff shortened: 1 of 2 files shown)


maxme:
Fix the problem! :shipit:

aforcier:
Note: `API>19` only, separate fix for toggling coming in a sparate PR

PR #327, opened 2016-03-23
Sync WPAndroid subtree and fix broken tests

aforcier:
Fixes unit tests broken by an unimplemented `EditorFragmentAbstract` method, as well as an integration test that broke because the `%%...%%` placeholders in `android-editor.html` weren't being replaced in the test.

cc @maxme


maxme:
:shipit:

Issue #120, opened 2015-04-06
Cursor in title field and keyboard up on new posts


bummytime:
On new posts, make sure the cursor is in the title field and keyboard is up ready to go! (For an example of this behavior, check out Evernote Android or WPiOS).

bummytime:
Also, lets do the same thing for existing posts.

thianhlu:
Has this been fixed? If not, title is not in focus upon new post/edit post.

aforcier:
@thianhlu there was a PR that addressed it but it ran into a lot of issues on API<19 so it was put on hold for the time being: https://github.com/wordpress-mobile/WordPress-Editor-Android/pull/126

bummytime:
From #235:

> Found yet another new related bug (sorry!). This one happens when visiting the home screen and then re-opening the app, under certain circumstances.
> 
> On a phone that doesn't rotate the home screen when in landscape mode:
> - Launch example app, and open the editor
> - In landscape mode, tap the content so that the keyboard appears and the action bar hides (this bug doesn't occur if the keyboard is down)
> - Go to home screen (it'll be in portrait mode, while you're still holding the phone in landscape - keep it like that)
> - Re-launch the editor
> - You'll see the editor screen, with no keyboard and no ActionBar
> 
> This seems not to happen on devices that rotate the home screen (which is all the tablets I tested, and for some reason API15 emulator).

rachelmcr:
A couple testers have noticed that the editor doesn't have focus when you start a new post. In addition, it causes some issues if you try to start adding content right away:
- Tap the media icon and you can select media to upload, but it isn't inserted into the post.
- Tap the link icon and you can enter a link and link text, but it isn't inserted into the post.
- You can tap the other formatting buttons and they are highlighted blue, but the editor doesn't gain focus. If you then tap into the editor and start typing, the button stays highlighted but isn't actually active.

Testers:
- Donncha, Galaxy S5, Marshmallow 6.0.1 (CM 13.0)
- Mo, Moto X 2, Android 6.0
- Me, Nexus 9, Android 6.0.1

aforcier:
I'm assigning this issue to the `1.0` (release) milestone. I think at the very least we can apply @bummytime's original fix for newer versions (`API19`+) for now, and get to older APIs in the future.

PR #329, opened 2016-03-24
Fix #302: DOM errors due to unfocused editor

### LINKED ISSUE #302: Fix DOM errors due to editor fields not in focus
Ref: https://github.com/wordpress-mobile/WordPress-Android/pull/3793#issuecomment-189059193

These errors are caused by the editor's fields not being in focus when the methods in question are called.

We should give the focus to the content field by default in such cases, or at least suppress the errors to debug status.

A common cause of this bug is the editor not being in focus on load, which is tracked in this issue: https://github.com/wordpress-mobile/WordPress-Editor-Android/issues/120.

A complete solution would involve forcing focus on load and preventing focus loss through HTML mode toggle/content deletion. In that case we can preserve errors of this type, since they'll be pointing us to new bugs.

Related issue: https://github.com/wordpress-mobile/WordPress-Editor-Android/issues/179

aforcier:
Fixes #302. With this PR we shouldn't get any more DOM errors in the log of the type:

``` javascript
Uncaught IndexSizeError: Failed to execute 'getRangeAt' on 'Selection': 0 is not a valid index
```

and

``` javascript
Uncaught Error: INDEX_SIZE_ERR: DOM Exception 1
```

from any API level (if you see any of these in testing let me know :wink: ).

The main cause of this bug was fixed as a sub-issue (#120). The remaining cases that I found were limited to `API<19`, and didn't appear to have any effect other than the log (later operations apparently gave focus where necessary and everything ran smoothly). It was still worth fixing them to avoid cluttering up crashlytics.

I also took the opportunity, since I was working on node traversal methods, to import this syntax typo fix from iOS: https://github.com/wordpress-mobile/WordPress-Editor-iOS/pull/774.

cc @maxme

### DIFF
--- a/libs/editor-common/assets/ZSSRichTextEditor.js
+++ b/libs/editor-common/assets/ZSSRichTextEditor.js
@@ -197,13 +197,18 @@ ZSSEditor.getFocusedField = function() {
 
     if (currentField) {
         currentFieldId = currentField.attr('id');
+    }
 
     while (currentField && (!currentFieldId || this.editableFields[currentFieldId] == null)) {
         currentField = this.closerParentNodeStartingAtNode('div', currentField);
         if (currentField) {
             currentFieldId = currentField.attr('id');
         }
+    }
 
+    if (!currentFieldId) {
+        ZSSEditor.resetSelectionOnField('zss_field_content');
+        currentFieldId = 'zss_field_content';
     }
 
     return this.editableFields[currentFieldId];
@@ -415,7 +420,6 @@ ZSSEditor.getCaretArguments = function() {
 };
 
 ZSSEditor.getJoinedFocusedFieldIdAndCaretArguments = function() {
-
     var joinedArguments = ZSSEditor.getJoinedCaretArguments();
     var idArgument = "id=" + ZSSEditor.getFocusedField().getNodeId();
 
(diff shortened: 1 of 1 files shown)


maxme:
Got this on a 4.1.1 emulator: 

`03-25 08:49:59.688 2633-2633/org.wordpress.android.beta E/WordPress-EDITOR: Uncaught Error: INDEX_SIZE_ERR: DOM Exception 1 -- From line 2782 of file:///android_asset/ZSSRichTextEditor.js`

maxme:
Should we check all calls to `window.getSelection()`?

aforcier:
It looks like the specific case you mentioned should be addressed (https://github.com/wordpress-mobile/WordPress-Editor-Android/pull/329/files#diff-cfce53bcfd107d0762d2395366d54762L2782), but I agree that we should check all calls to `window.getSelection()`, even if they're not causing any issues that we know of yet.

Luckily `window.getSelection()` returns an empty Selection object rather than null if there's no selection, so things like `selection.rangeCount` or `selection.removeAllRanges()` don't cause any problems.

aforcier:
Added checks for all unsafe calls to `window.getSelection()` in https://github.com/wordpress-mobile/WordPress-Editor-Android/pull/329/commits/400a5f72bda12f71ff855468e8aa47d54f0262cd.

maxme:
:shipit: - That's actually great to log these errors in Crashlytics, we'll see if we still have them in alpha-5+

PR #330, opened 2016-03-25
Issue #299: Allow double Enter to exit blockquotes

### LINKED ISSUE #299: Can't exit blockquote
I want to add a new paragraph after a blockquote but I don't seem to be able to? The only solution is to switch to the HTML editor.

aforcier:
Fixes #299. Brings the new editor to parity with the web editor by allowing the user to exit a blockquote by pressing Enter twice.

The pre-existing way of pressing the blockquote button in the format bar still works.

cc @maxme


maxme:
Looks good, I'm seeing weird things on an API 16 emulator when I switch to HTML mode and back, but it's probably not related to this PR.

:shipit:

PR #332, opened 2016-03-31
Fix #307: Expand selection to the word when creating a link and nothing is selected

### LINKED ISSUE #307: If cursor is in the middle of a word, pass the word to the Link dialog as the link text argument
From discussion around #268 (also involving PR #306)

Relevant text from the Issue:

> After toying around I think it would be nice if the link text field would auto-fill with whatever text your cursor is in the middle of. Currently it will only fill that field if text is highlighted.

Relevant text from the PR:

> if I've typed 'hello world' and put my cursor somewhere inside 'world' (with nothing highlighted) it should pre-fill 'world'

Where to implement:

> It looks like that will have to be done JS side in getSelectedText as well as in the EditorFragment.

maxme:
Fix #307: Expand selection to the word when creating a link and nothing is selected

### DIFF
--- a/WordPressEditor/build.gradle
+++ b/WordPressEditor/build.gradle
@@ -3,7 +3,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.0.0-beta7'
+        classpath 'com.android.tools.build:gradle:2.0.0-rc1'
     }
 }
 
--- a/WordPressEditor/build.gradle
+++ b/WordPressEditor/build.gradle
@@ -23,8 +23,8 @@ android {
     buildToolsVersion "23.0.2"
 
     defaultConfig {
-        versionCode 6
-        versionName "0.6"
+        versionCode 8
+        versionName "0.8"
         minSdkVersion 14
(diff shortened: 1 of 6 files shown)


aforcier:
This works great when inside an actual word, but if you're on a new line or after a space character, it targets the previous word: https://cloudup.com/cDh3NKtENTN

maxme:
Sadly, we can't use `selection.modify` because of this bug.

maxme:
Updated with some dirty code to grow auto select the word behind the caret. I wonder if we should disable this when the caret is at the beginning or end of a word.

aforcier:
Looks good @maxme - I think we can leave it as is, it seems more natural to me that the caret should behave the same way at the beginning/end of a word as in the middle somewhere.

Found one issue on API 16 but it's not caused by this PR - it looks like pressing the link button with selected text doesn't add the selected text to the dialog. The same happens with your changes if the caret is inside a word. I'll open a new issue for that.

:shipit:

aforcier:
Opened issue for API16 bug: https://github.com/wordpress-mobile/WordPress-Editor-Android/issues/339

PR #333, opened 2016-04-01
fix #258 ask to remove failed uploads when switching to html mode

### LINKED ISSUE #258: Refine HTML mode toggle behavior while post contains failed image uploads
Currently, it's possible to switch to HTML mode while there are failed image uploads in the post, allowing you to see the overlay container and progress bar HTML:

failed-image-html

It might be better to instead show a dialog, blocking access to HTML mode unless all failed images are cleared. Here's how that looks on iOS:

ios-editor-html-mode-failed-media

maxme:
fix #258 ask to remove failed uploads when switching to html mode

### DIFF
--- a/WordPressEditor/build.gradle
+++ b/WordPressEditor/build.gradle
@@ -3,7 +3,7 @@ buildscript {
         jcenter()
     }
     dependencies {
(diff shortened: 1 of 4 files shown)


aforcier:
Working well on all API levels and when I tried it in WPAndroid. :shipit:

PR #334, opened 2016-04-04
Issue #292: Buggy autocorrect when using backspace on API 19+

### LINKED ISSUE #230: First character inserted appears as zw space
repro
- on a nifty kitkat samsung device, tap create post
- type title
- insert caret in body
- tap any letter
- see zero width whitespace inserted instead of capital letter

### LINKED ISSUE #292: Backspacing text that includes whitespace characters can move line text and/or cursor positioning
[See video.](https://cldup.com/KInnNYGw0d.mp4)

Tough to describe...rather than a bunch of run-on sentences I'll put some informative bullets.

Problem occurs:
- when line is not first line of text
- first character on line must not be whitespace (a space character, for instance)
- must add a space character to end of line
- after backspacing the space at end of line _then_ backspacing a non-whitespace character

When it reproduces you'll notice the cursor may or may not be positioned incorrectly in the preceding word.

The problem line may or may not jump to the line above, looks as if you deleted a newline character.

Characters in the word you are backspacing may or may not duplicate at cursor position or at the front of the word. For example, backspacing `Test` may result in `TestTesst` on the previous line.

Entering HTML editing then returning guarantees that it will not reproduce within current text.

aforcier:
Fixes #292 (word in progress).

Uses `div` instead of `p` as the paragraph separator within the editor. Not yet complete: blockquotes are badly broken and further testing is needed on older APIs and in edge cases. This PR introduces a pretty fundamental change to editor internals, and could have any number of side effects. We'll need to test thoroughly before merging this branch, even once all current known issues are fixed.

<h4>TODO</h4>
- [x] Fix blockquotes, they're pretty badly broken by these changes
- [x] Fully test lists and all other formatting
- [x] Check older APIs throughly - if we run into issues we may need to limit these changes to `API 19+` and keep using `p` separators for older APIs (the bug only occurs on `API 19+` anyway)
- [x] Check Android N (with its beta WebView)
- [x] Make sure hardware keyboards work correctly
  - [x] Emulators
  - [x] Marshmallow devices
  - [x] Pre-API19 devices
  - [x] Android N
- [x] Make sure the `ZSSEditor.getFocusedField()` simplification doesn't break anything
- [x] Investigate zero-width space behavior: it seemed safe to remove it (and thus fix #230) when using `p` tags, but I haven't been able to get rid of zero-width space when using `div` tags without causing issues. While they're not an issue for `API 22+`, #230 is a pretty bad bug for Kit Kat users. It's not clear that we should implement the fix in this PR for KitKat, if it means blocking us from fixing #230. This needs some further thought/research.

<h4>Details/backstory</h4>

Some false positive results at first suggested that the issue was caused by the zero-width space character [we're using](https://github.com/wordpress-mobile/WordPress-Editor-Android/blob/5e905787de707b4c43dc6cadedebbd315980bf35/libs/editor-common/assets/ZSSRichTextEditor.js#L3256), which is also the cause of [this bug](https://github.com/wordpress-mobile/WordPress-Editor-Android/issues/230) on API 19-21. That turns out not to be the case. The takeaway is to vary the test words/sequences used when testing, because autocorrect can learn your sequence and stop trying to correct it when you backspace, giving a false positive result when in fact the bug is still present.

The real culprit is a bug in the Android WebView (or the autocorrect/software keyboard, or all of them). By default, paragraphs are separated using divs in the WebView, but this was changed in the editor from iOS to fix some paragraph styling/caret issues: https://github.com/wordpress-mobile/WordPress-Editor-iOS/pull/319.

The change was made using:

``` javascript
document.execCommand('defaultParagraphSeparator', false, 'p');
```

While this method is implemented on Android, it causes issues with autocorrect, at least when using the default keyboard. I opened a ticket in the Chromium project about this, with a basic demo project showcasing the bug. If a fix comes out for this in the future we may be able to revert this PR (if we even merge it), and keep `p` tags.

Chromium project ticket: https://bugs.chromium.org/p/chromium/issues/detail?id=599890

### DIFF
--- a/libs/editor-common/assets/ZSSRichTextEditor.js
+++ b/libs/editor-common/assets/ZSSRichTextEditor.js
@@ -474,7 +474,7 @@ ZSSEditor.getYCaretInfo = function() {
     //
     if (needsToWorkAroundNewlineBug) {
         var closerParentNode = ZSSEditor.closerParentNode();
-        var closerDiv = ZSSEditor.closerParentNodeWithName('div');
+        var closerDiv = ZSSEditor.findParentContenteditableDiv();
 
         var fontSize = $(closerParentNode).css('font-size');
         var lineHeight = Math.floor(parseInt(fontSize.replace('px','')) * 1.5);
--- a/libs/editor-common/assets/ZSSRichTextEditor.js
+++ b/libs/editor-common/assets/ZSSRichTextEditor.js
@@ -191,26 +191,7 @@ ZSSEditor.getField = function(fieldId) {
 };
 
 ZSSEditor.getFocusedField = function() {
-    var currentField = $(this.closerParentNodeWithName('div'));
-    var currentFieldId;
-
-    if (currentField) {
-        currentFieldId = currentField.attr('id');
-    }
-
-    while (currentField && (!currentFieldId || this.editableFields[currentFieldId] == null)) {
-        currentField = this.closerParentNodeStartingAtNode('div', currentField);
-        if (currentField) {
-            currentFieldId = currentField.attr('id');
-        }
-    }
-
-    if (!currentFieldId) {
-        ZSSEditor.resetSelectionOnField('zss_field_content');
-        currentFieldId = 'zss_field_content';
-    }
-
-    return this.editableFields[currentFieldId];
+    return this.focusedField;
 };
 
 ZSSEditor.execFunctionForResult = function(methodName) {
@@ -2294,7 +2275,7 @@ ZSSEditor.removeVisualFormatting = function( html ) {
     str = ZSSEditor.replaceVideoPressVideosForShortcode( str );
     str = ZSSEditor.replaceVideosForShortcode( str );
     return str;
-}
+};
 
 ZSSEditor.insertHTML = function(html) {
 	document.execCommand('insertHTML', false, html);
--- a/libs/editor-common/assets/ZSSRichTextEditor.js
+++ b/libs/editor-common/assets/ZSSRichTextEditor.js
@@ -63,7 +63,7 @@ ZSSEditor.editableFields = {};
 ZSSEditor.lastTappedNode = null;
 
 // The default paragraph separator
-ZSSEditor.defaultParagraphSeparator = 'p';
+ZSSEditor.defaultParagraphSeparator = 'div';
 
 // Video format tags supported by the [video] shortcode: https://codex.wordpress.org/Video_Shortcode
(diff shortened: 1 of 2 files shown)


maxme:
My first testing session went pretty well (Sony Z5C), seems to fix #292 when suddendly I experienced this: https://cloudup.com/cxY8HPO0VXs (weird things happen around 19s - not sure if it's related to the use of autocorrect / word prediction). IMO this issue is minor compared to #292 because 1. I wasn't able to repro it consistently, 2. changes are made in front of the cursor when you're deleting, it's easy to delete one more character.

maxme:
Sony Z5C: https://cloudup.com/ceBmWZVR91W weird problem when deleting text around 30s in the video. Emojis not related (I reproduced this issue with A / B / C chars).

aforcier:
Updated.
- Fixed broken blockquotes (seems to work as normal on all API levels now)
- Fixed an API<19 issue where deleting the only character in a post would cause the cursor to lose focus (DOM tree wasn't being parsed correctly, it needed to account for `div` as paragraph separators)

aforcier:
Ready for full review @maxme.

Remaining issues:
- The ZW space issue affects KitKat fairly badly, and to a lesser extent API 21+ with hardware keyboard (if you hold down shift at the beginning of the post for capitalization, the zw character appears; however, it's highlighted, and as soon as you press a key, the letter you pressed appears and replaces the zw character). I'll start researching that more thoroughly, but there's already an open issue and if part of this PR needs to be reverted for API19, we'll do that as part of the PR for that issue.
- On my Samsung S3 API18, the keyboard isn't automatically capitalized after hitting enter anymore. The emulators for that API level don't capitalize at all, but that's how they've always behaved (unsure if this is an emulator issue or not). I'd like to test this on the API16 I'll have soon, and also on new Samsung devices (such as the S6).

maxme:
<strike>Hardware keyboard issues (Sony Z5C, android 6.0):
1. Open the editor, type `TAB` to give the focus to the content, type `TAB` again, see the ZW char `&#x200b` in content
2. Open the editor, type `TAB` twice (get a `&#x200b`), focus is given to the button bar but when you type, it adds char to the content field (like the content field had focus). This is weird because the cursor is invisible, and there is no visual feedback about current focus.
   </strike>

Not related to this PR, moved this comment to https://github.com/wordpress-mobile/WordPress-Editor-Android/issues/230

maxme:
:100: :+1: :shipit:

PR #337, opened 2016-04-08
Issue #336: Fix paragraph layout when inserting media

### LINKED ISSUE #336: Images aren't added in their own paragraph
#334 introduced a bug where adding an image doesn't add a new div after at, but instead you remain inside the image's div. This causes, e.g., successive images to be added as increasingly nested div tags within the first image's div.

aforcier:
Fixes #336, changing paragraph handling to work around the WebView's strictness with divs. The bug was introduced by the change from `p` tags to `div` tags in https://github.com/wordpress-mobile/WordPress-Editor-Android/pull/334.

This change should keep text added after images in a separate paragraph at the same DOM hierarchy level, rather than nesting them inside the image's own paragraph (also applies to video, and galleries).

cc @maxme

### DIFF
--- a/libs/editor-common/assets/ZSSRichTextEditor.js
+++ b/libs/editor-common/assets/ZSSRichTextEditor.js
@@ -740,12 +740,12 @@ ZSSEditor.setBackgroundColor = function(color) {
 };
 
 /**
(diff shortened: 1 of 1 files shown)


maxme:
:shipit:

PR #342, opened 2016-04-09
Fix #340: Wrap links in a new paragraph when the post is empty

### LINKED ISSUE #340: Links at start of post should be wrapped in paragraph tags
It seems we never wrapped links added to new posts in paragraph tags. This seems to cause some issues on my Samsung S3 after the switch from `p` to `div` tags.

aforcier:
Fixes #340, wrapping links in paragraph tags when they're added to an empty post.

cc @maxme


maxme:
:shipit: (Confirmed by raw HTML debug print).

Issue #290, opened 2016-02-23
Add optimistic upload progress for media


aforcier:
When uploading videos (and sometimes images), especially on slower connections, a while may pass before the progress bar is updated, and it's not clear to the user whether anything is actually happening.

We can add handling that will update the progress bar automatically, regardless of actual upload progress, up to ~15%. This will slow down and eventually stop, and wait for the real upload to catch up.

Original ref: https://github.com/wordpress-mobile/WordPress-Android/pull/3773#issuecomment-186860820

aforcier:
It would also be good to do the same for the end of uploads; sometimes (due to the nature of the progress relying on bytes uploaded), on slow connections, the upload can appear to be stuck with a full progress bar for ~30 seconds.

We should probably just ignore progress updates past ~90%, so the bar never actually completes, and is simply removed at 90% whenever the upload is really complete.

aforcier:
Merged in #343.

PR #343, opened 2016-04-11
Feature #290: Optimistic upload progress

aforcier:
Addresses #290. Adds 'optimistic' upload progress support for media uploads.

The progress bar will gradually increment to 15%, increasingly slowly, until it either stops at 15% (and waits for the real upload progress to catch up) or is overtaken before then by the real upload progress.

This PR also limits the displayed progress to 90%. This prevent situations where the progress shows full but there is still processing going on, and the upload appears to have bugged out. Now, the progress bar will remain at 90% until the upload is really complete and the local media url is replaced with the uploaded url.

cc @maxme


maxme:
:shipit:

Issue #309, opened 2016-03-04
Selected text is sometimes ignored by link dialog fragment


aforcier:
Originally reported here: https://github.com/wordpress-mobile/WordPress-Editor-Android/pull/306#issuecomment-191282968

demo: https://cloudup.com/crGEVVRSy2t

The selected text is ignored by the link dialog fragment, and if the "Link text" field is left blank, the selected text is deleted.

This bug has only been confirmed on emulators so far (API `16`, `21`). The bug doesn't occur on API `18`, `22` or `23` emulators or devices.

The range of affected versions needs to be investigated, and we should make sure this isn't an emulator-only issue.

aforcier:
Confirmed to occur on physical API16 device (Nexus S).

PR #351, opened 2016-04-14
Fix #347: Fix content loss on slower/older devices

### LINKED ISSUE #347: Content loss when switching to HTML mode on older/slower devices
Possibly the only cause of content loss tracked in https://github.com/wordpress-mobile/WordPress-Editor-Android/issues/47. This is caused by a race condition: the `CountDownLatch` `await` inside `getTitle()` and `getContent()` seems to be holding up the editor from receiving the title/content strings from JavaScript.

Confirmed on Nexus S API16, and on several emulators.

aforcier:
Fixes #347 and #309.

There was a race condition problem in the way we were retrieving content from the ZSSEditor. `ZSSEditor.execFunctionForResult()` executes successfully, but the request never reaches the ZSSEditor until the UI thread is unblocked (when the `CountDownLatch` wait times out). I'm not entirely sure why this is happening, as the JS runs on its own thread, but this is likely due to some older `WebView` internals. The issue consistently occurred on the (very slow) API16 Nexus S and with some frequency on emulators of various API levels.

I solved the issue by moving `CountDownLatch`-reliant processes to worker threads, with the worker thread responsible for updating the UI when the JS requests return.

### DIFF
--- a/WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragment.java
+++ b/WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragment.java
@@ -517,21 +517,38 @@ public class EditorFragment extends EditorFragmentAbstract implements View.OnCli
 
             String selectedText = mSourceViewContent.getText().toString().substring(mSelectionStart, mSelectionEnd);
             dialogBundle.putString(LinkDialogFragment.LINK_DIALOG_ARG_TEXT, selectedText);
+
+            linkDialogFragment.setArguments(dialogBundle);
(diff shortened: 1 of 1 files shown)


aforcier:
Note: Another (smaller) content loss case can happen when publishing from WPAndroid with changes since the last autosave.

Opened a separate issue: https://github.com/wordpress-mobile/WordPress-Editor-Android/issues/352.

aforcier:
cc @maxme

maxme:
:shipit:

Issue #315, opened 2016-03-08
Pasting text and editing it results in extra lines


rachelmcr:
This looks to be the same as or very similar to https://github.com/wordpress-mobile/WordPress-Editor-iOS/issues/783 in the iOS visual editor. cc @hoverduck who mentioned the issue in the Android visual editor, which led to my testing:

**Steps to reproduce:**
1. Copy a block of text and paste into the editor
2. Select a word in the text
3. Backspace (to try to delete the word) or type something (to try to overwrite the word)

Result: Lines breaks are added to the text. If you backspace the selected word, the first letter of that word isn't erased and the new line starts with the following word. If you type something, your typing starts on the new line. Trying to backspace again will result in more new lines added.

I reproduced this each time I copied and pasted text into the editor, no matter the source (including when I copied text that I had previously typed directly into the editor and then pasted it into the editor again).

Tested on Nexus 9, Android 6.0.1, app version alpha-2.

kwonye:
Repro'd on Nexus 6P, Android 6.0.1, version alpha-4 as well

aforcier:
From my testing this seems to only happen when the editor is empty (the pasted text is not wrapped in paragraph tags, causing the editing bugs).

However, I am seeing that formatting will be kept if you cut and paste text within the editor, which will sometimes nest the pasted text inside existing paragraph tags (instead of pasting them at the top layer).

Issue #257, opened 2015-10-07
Buggy behavior when pasting into an empty post with hardware keyboard on old APIs (ZW space-related)


aforcier:
4.1.1 emulator, HW keyboard:
- Paste string into content field
- Try to delete it (by pressing delete key on hardware keyboard)
- Deleting fails
- Use up/down arrow keys to navigage
- Editor keeps adding `&amp;#x200b;` into the field while spitting this log:

```
10-06 13:48:49.332 7910-7910/org.wordpress.android D/WordPress-EDITOR: Uncaught Error: INDEX_SIZE_ERR: DOM Exception 1 -- From line 2051 of file:///android_asset/ZSSRichTextEditor.js
10-06 13:48:49.496 7910-7910/org.wordpress.android D/WordPress-EDITOR: Uncaught Error: INDEX_SIZE_ERR: DOM Exception 1 -- From line 2051 of file:///android_asset/ZSSRichTextEditor.js
10-06 13:48:49.652 7910-7910/org.wordpress.android D/WordPress-EDITOR: Uncaught Error: INDEX_SIZE_ERR: DOM Exception 1 -- From line 2051 of file:///android_asset/ZSSRichTextEditor.js
10-06 13:48:49.808 7910-7910/org.wordpress.android D/WordPress-EDITOR: Uncaught Error: INDEX_SIZE_ERR: DOM Exception 1 -- From line 2051 of file:///android_asset/ZSSRichTextEditor.js
10-06 13:48:50.200 7910-8370/org.wordpress.android D/WordPress-EDITOR: callback-response-string: function=getHTMLForCallback~id=zss_field_title~contents=
10-06 13:48:50.204 7910-8370/org.wordpress.android D/WordPress-EDITOR: callback-response-string: function=getHTMLForCallback~id=zss_field_content~contents=INDEX_SIZE_ERR: DOM Exception 1 -- From line 2051 of file:///android_asset/ZSSRichTextEditor.js
10-06 13:48:50.244 7910-7914/org.wordpress.android D/dalvikvm: GC_CONCURRENT freed 1510K, 17% free 20703K/24711K, paused 12ms+2ms, total 35ms
10-06 13:48:50.908 7910-7976/org.wordpress.android W/DebugConnManager: getNetworkInfo() on networkType 1
[...]

PR #354, opened 2016-04-15
Issue #230: First character appears as zero-width space

### LINKED ISSUE #230: First character inserted appears as zw space

aforcier:
Fixes #230.

There are two separate bugs fixed by this PR. They are both caused by our use of the `&#x200b;` character (zero-width space) when we add a new paragraph. We rely on this to force the cursor to move inside the `<div></div>` tag we add. It was previously possible to avoid using it when `p` was the paragraph separator, but since we changed that to `div` in order to work around the backspace autocorrect bug (https://github.com/wordpress-mobile/WordPress-Editor-Android/pull/334), it's no longer possible. The WebView seems to be stricter about `div` tags than `p` tags for some reason.

<h4>1. Hardware keyboard non-printable keycodes</h4>

On hardware keyboards at all API levels, entering non-printable keycodes (tab, ctrl, shift, etc.) were still triggering the paragraph insertion code. When a normal key is entered, it instantaneously replaces the filler `&#x200b;` space with the character entered and nothing odd is visible to the user. Non-printable keycodes, however, just caused the space character to be entered, and it would stay there until a printable key was pressed to replace it.

We're now ignoring non-printable characters from hardware keyboards in the new paragraph handling code, so that paragraphs aren't added until a real character is entered.

<h4>2. API19 doesn't like zero-width space in new paragraphs</h4>

On an `API19` software keyboard, the first character pressed is always replaced by the `&#x200b;` literal. If the user keeps typing it will go away, but they will lose the first character they entered. I'm not sure why this is happening, but it's a WebView issue that was introduced at version `37` and fixed by version `39`. This means that it affects all `API19` devices, and also `API21` devices running the default, never updated Android System WebView.

The fix for this was to stop wrapping the first text entered in paragraph tags on `API19`, and instead correct it when the user enters a new line. This isn't ideal, but so far I haven't found any side effects (or a better solution). All the same, I've limited the change to `API19` only to avoid possibly breaking things for other API levels that didn't even need this fix.

I've considered extending the fix to `API21` as well. Unless a device's OEM has disabled the standalone Android System WebView app, any `API21` device that has updated their WebView app since ~early 2015 should not have the bug anymore. I've tried to discover if there are any OEMs who have done this; Samsung (usually the worst case scenario for WebView customization by OEM) at least hasn't, or at least not on the S6.

Instead, we'll [track](https://github.com/wordpress-mobile/WordPress-Android/issues/3978) the WebView version in our analytics, so we can look for users running version `<39` of the Android System WebView on `API21` and extend the fix accordingly.

Note that stock `API22` ships with `v39`, which doesn't exhibit this bug.

cc @maxme

### DIFF
--- a/libs/editor-common/assets/ZSSRichTextEditor.js
+++ b/libs/editor-common/assets/ZSSRichTextEditor.js
@@ -556,12 +556,6 @@ ZSSEditor.getYCaretInfo = function() {
     return this.caretInfo;
 };
 
-// MARK: - Default paragraph separator
-
-ZSSEditor.defaultParagraphSeparatorTag = function() {
-    return '<' + this.defaultParagraphSeparator + '>';
-};
-
 // MARK: - Styles
 
 ZSSEditor.setBold = function() {
@@ -684,9 +678,9 @@ ZSSEditor.setHeading = function(heading) {
 	var formatBlock = document.queryCommandValue('formatBlock');
 
 	if (formatBlock.length > 0 && formatBlock.toLowerCase() == formatTag) {
-		document.execCommand('formatBlock', false, this.defaultParagraphSeparatorTag());
+		document.execCommand('formatBlock', false, Util.buildOpeningTag(this.defaultParagraphSeparator));
 	} else {
-		document.execCommand('formatBlock', false, '<' + formatTag + '>');
+		document.execCommand('formatBlock', false, Util.buildOpeningTag(formatTag));
 	}
 
 	ZSSEditor.sendEnabledStyles();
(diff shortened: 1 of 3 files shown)


aforcier:
Marking this as 'Do Not Merge' for now - my findings from working on https://github.com/wordpress-mobile/WordPress-Editor-Android/issues/315 and https://github.com/wordpress-mobile/WordPress-Editor-Android/issues/257 are that with some modifications this PR can fix those issues too, but in its present state it makes certain things worse.

I'm going to try and make it all work at once, if it gets out of hand I'll address them separately.

aforcier:
After further examination, it's probably better to keep the pasting fixes separate.

The main change affecting this PR will probably be dropping the API 19 restriction here: https://github.com/wordpress-mobile/WordPress-Editor-Android/pull/354/files#diff-cfce53bcfd107d0762d2395366d54762R3094 (or possibly loosening it to `API > 19` instead of `API != 19`). In that case, I think we can test/merge this first, and just retest ZW space on API19 within the PR for #315 and #257.

maxme:
Works great, tested on 
- [x] Nexus 9 (Android N - preview 2)
- [x] Z5C (Android 6)
- [x] Geny API 19 (virtual keyboard disabled)
- [x] Geny API 16 (virtual keyboard disabled)
- [x] Geny API 21 (virtual keyboard disabled)

There is still one character that adds a zw space: backquote ("grave accent") - seems like this is the only one that could be an issue from this list https://css-tricks.com/snippets/javascript/javascript-keycodes/

<s>Not introduced by this PR, but I noticed with a hardware keyboard configured in French, you can't type accented characters in the editor.</s> (I was wrong, it works fine).

aforcier:
Added 'grave accent' to the list of ignored modifier characters.

maxme:
:shipit:

PR #356, opened 2016-04-17
Issue #353: HTML entities in title field

### LINKED ISSUE #353: HTML entities in title field after switching between visual and HTML views
Special characters in the title field (such as `&` or `>`) are encoded as HTML entities when you switch to HTML view. Switching back to visual retains those entities, and they are re-encoded every time you switch. (So `&` becomes `&amp;`, and if you switch again it becomes `&amp;amp;`, and so on.)
### Steps to reproduce
1. Open the app and start a new post
2. Add a title that includes a special character like `&`
3. Switch to HTML view and note the HTML entity in the title
4. Switch back to visual view and note that the HTML entity persists

You can also see that HTML entity in the post list and on the published post, although the local preview converts the entity to the character.
### Screenshots

Before, in visual:

After switch to html and back to visual:

aforcier:
Fixes #353 by returning plaintext when obtaining the title field's contents from the ZSSEditor.


maxme:
:shipit:

PR #357, opened 2016-04-21
Issue/350 fix newlines api16

maxme:
Due to the introduction of the pseudo-callback receiver `WebViewClient.shouldOverrideUrlLoading` in this PR  https://github.com/wordpress-mobile/WordPress-Editor-Android/pull/289 for API < 17. We had an issue with HTML, that was somehow encoded and stripped. Using encoderURI before calling the callback and decodeURI when receiving the data ensure the string data won't be automatically changed/stripped.

This PR should only affect API <= 16


aforcier:
Nice fix @maxme, working well on my `API16` Nexus S as well as the emulator. :shipit:

PR #358, opened 2016-04-21
Issue #315: Buggy behavior when pasting text in empty posts

### LINKED ISSUE #315: Pasting text and editing it results in extra lines

aforcier:
Fixes #315 and #257.

A few separate things were addressed in this PR.
1. When software pasting (long press -> paste) in an empty post, the paste wasn't being wrapped in paragraph tags
2. Same as above, but for hardware keyboard pasting (`ctrl` + `v`)
3. Fixed two issues with blockquotes, partially caused by the above changes. When the first line of a post was wrapped in blockquotes, or a post was started with a blockquote, the blockquote contents weren't wrapped in paragraph tags. This caused formatting as well as autocorrect issues. This was most likely introduced in https://github.com/wordpress-mobile/WordPress-Editor-Android/pull/354 by [this change](https://github.com/wordpress-mobile/WordPress-Editor-Android/pull/354/commits/d066fbe8ac22761d26dd73329b55f45563395a4f#diff-cfce53bcfd107d0762d2395366d54762R3094), but only for API19. To make the paste fixes for this PR that line changed to apply at all API levels, and I noticed the bug on my 5X.

cc @maxme

### DIFF
--- a/libs/editor-common/assets/ZSSRichTextEditor.js
+++ b/libs/editor-common/assets/ZSSRichTextEditor.js
@@ -3024,6 +3024,9 @@ ZSSField.prototype.bindListeners = function() {
     this.wrappedObject.bind('input', function(e) { thisObj.handleInputEvent(e); });
     this.wrappedObject.bind('compositionstart', function(e) { thisObj.handleCompositionStartEvent(e); });
     this.wrappedObject.bind('compositionend', function(e) { thisObj.handleCompositionEndEvent(e); });
+
+    // Only supported on API19+
+    this.wrappedObject.on('paste', function(e) { thisObj.handlePasteEvent(e); });
 };
 
 // MARK: - Emptying the field when it should be, well... empty (HTML madness)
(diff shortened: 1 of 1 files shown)


maxme:
:shipit: confirmed fixes for #257 and #315

PR #359, opened 2016-04-21
Fix #279: hint for users to enable Enhanced Web Accessibility if they need it

### LINKED ISSUE #279: Help users of the new editor whose setup isn't working with the visual editor
Help users of the new editor whose setup isn't working with the visual editor (right now this looks like _Explore by Touch_ + software keyboard, but there are likely other problematic cases). The overall page navigation works, so maybe we can add a way to change to the legacy editor, or default to HTML mode? Otherwise, some sort of prompt, something lsuggested here: https://github.com/wordpress-mobile/WordPress-Editor-iOS/issues/608#issuecomment-115333537

Apparently there is a way to check [touch exploration is enabled](https://developer.android.com/reference/android/view/accessibility/AccessibilityManager.html#isTouchExplorationEnabled%28%29), we could prompt the user here (and remember their preference).

Ref: #183

maxme:
Fix #279: hint for users to enable Enhanced Web Accessibility if they need it

I was hesitant to add this on wpandroid side.

### DIFF
--- a/WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragment.java
+++ b/WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragment.java
@@ -7,6 +7,7 @@ import android.content.Context;
 import android.content.DialogInterface;
 import android.content.DialogInterface.OnClickListener;
 import android.content.Intent;
+import android.content.SharedPreferences;
 import android.content.res.Configuration;
 import android.net.Uri;
 import android.os.Build;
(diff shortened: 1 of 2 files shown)


aforcier:
Looks like 'Enhanced web accessibility' was [dropped](https://www.paciellogroup.com/blog/2015/10/accessibility-testing-with-android-talkback/) as of `4.2`.

As discussed, closing this PR, we'll have to find another solution. :disappointed:

PR #360, opened 2016-04-26
Issue #344: Unable to backspace to remove blockquotes

### LINKED ISSUE #344: Removing all content from a blockquote does not remove blockquote tags
If you create a blockquote with text, and then backspace to remove all of the blockquote text (up to the previous line) the blockquote HTML tags and styling persist.

The only exception is if you remove all of the post content, at which point the blockquote is removed.

Screencast: https://cloudup.com/cKekZu480HF

Tested with Nexus 9, Android 6.0.1, app v. alpha-6.

aforcier:
Fixes #344.

Backspacing the entire contents of a blockquote was clearing the `div` tags inside it, causing backspace to be unable to delete it (WebView shenanigans). This also meant that when typing content back in, we would sometimes end up with bad autocorrect bugs.

The 'usual' way of inserting `&#x200b;` characters to force the div tag to stay open didn't work for blockquotes.

The only fix I found was to add a 'afterTextChanged' feature, which is fired with a delay whenever input is detected. This seemed to work okay, even on API16, but just in case I added a check to avoid running the method if the HTML hasn't actually been modified (which might mean `afterKeyDownEvent` was called too soon): https://github.com/wordpress-mobile/WordPress-Editor-Android/commit/2206af5203a43a70288c2ebff9bcf68900714338

cc @maxme

### DIFF
--- a/libs/editor-common/assets/ZSSRichTextEditor.js
+++ b/libs/editor-common/assets/ZSSRichTextEditor.js
@@ -3305,6 +3305,13 @@ ZSSField.prototype.afterKeyDownEvent = function(e) {
         newRange.setEnd(newFocusEle, 1);
         selection.removeAllRanges();
         selection.addRange(newRange);
+    } else if (focusedNode.nodeName == NodeName.DIV && focusedNode.parentNode.nodeName == NodeName.BLOCKQUOTE
+        && focusedNode.parentNode.childNodes.length == 1
+        && (focusedNode.innerHTML.length == 0 || focusedNode.innerHTML == '<br>')) {
+        // When a post begins with a blockquote, and there's content after that blockquote, backspacing inside that
+        // blockquote will work until the blockquote is empty. After that, backspace will have no effect
+        // This fix identifies that situation and makes the call to setBlockquote() to toggle off the blockquote
+        ZSSEditor.setBlockquote();
     }
 };
 
--- a/libs/editor-common/assets/ZSSRichTextEditor.js
+++ b/libs/editor-common/assets/ZSSRichTextEditor.js
@@ -3292,9 +3292,7 @@ ZSSField.prototype.handlePasteEvent = function(e) {
  *  @brief      Fires after 'keydown' events, when the field contents have already been modified
  */
 ZSSField.prototype.afterKeyDownEvent = function(beforeHTML, e) {
-    if (beforeHTML == e.target.innerHTML) {
-        return;
-    }
+    var htmlWasModified = (beforeHTML != e.target.innerHTML);
 
     var selection = document.getSelection();
     var range = selection.getRangeAt(0).cloneRange();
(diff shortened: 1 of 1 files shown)


aforcier:
I discovered that one of my fixes (backspace to delete empty blockquote at the start of a post, with more content below it) was broken by the modified HTML check (https://github.com/wordpress-mobile/WordPress-Editor-Android/commit/2206af5203a43a70288c2ebff9bcf68900714338).

I changed that in https://github.com/wordpress-mobile/WordPress-Editor-Android/pull/360/commits/ed096e24997f6b688af356995594cf58fe39408f.

Now everything works, but if the timeout delay isn't enough (slow device maybe), blockquotes will be turned off after entering the first character inside one.

On a new line, insert a blockquote:

``` html
<div>Text</div>
<blockquote>
    <div><br></div>
</blockquote>
```

Type a character inside the blockquote:

``` html
<div>Text</div>
<div>A</div>
```

Turning the blockquote back on and typing at that point will work fine, since the blockquote wasn't empty when we started typing. I haven't been able to reproduce this yet (I was able to force it by calling `afterKeyDownEvent` with no delay), but it's not impossible. I'll try to mitigate the problem a bit by adding another check [here](https://github.com/wordpress-mobile/WordPress-Editor-Android/pull/360/files#diff-cfce53bcfd107d0762d2395366d54762R3319), so that it also checks that the blockquote is the first element in the post (which is the only situation where the bug happens).

Edit: Added the 'parent blockquote is first element in document' check in https://github.com/wordpress-mobile/WordPress-Editor-Android/pull/360/commits/a5b021d7b6be0cf462dcdb4bd048dfc3ae00fd7e.

aforcier:
Note: The original issue (backspace inside any empty blockquote doesn't remove the blockquote) is unfortunately not fixed on `API<19` when using software keyboards. The reason for this is that when in that situation, `keydown` and `input` jQuery events aren't being fired at all when pressing backspace.

Since the user can still remove the blockquote by untoggling the format bar button, and the worse issue of autocorrect bugs doesn't occur below API19, my recommendation is to just ignore this on those API levels.

The second fix (backspace to remove blockquotes at the start of a post, with content below it) still works and is still required on `API<19`.

maxme:
- [x] API 16+hardware keyboard - Removing all content from a blockquote at start of a post remove blockquote tag (Behavior B)
- [x] API 16+hardware keyboard  - Removing all content from a blockquote at start of a post with content below it remove blockquote tag (Behavior B)
- [x] API 16+hardware keyboard  - Removing all content from a blockquote remove blockquote tag (Behavior A2)
- [x] API 23 - Removing all content from a blockquote at start of a post remove blockquote tag (Behavior B)
- [x] API 23 - Removing all content from a blockquote at start of a post with content below it remove blockquote tag (Behavior A)
- [x] API 23 - Removing all content from a blockquote remove blockquote tag (Behavior A)

**Behavior A**: when the user type backspace and the blockquote is empty
**Behavior A2**: when the user type backspace and the blockquote is empty, and the character before the caret is not a newline
**Behavior B**: when the last character in the blockquote is deleted

maxme:
:shipit:

PR #363, opened 2016-04-28
Issue #345: Buggy links inside blockquotes

### LINKED ISSUE #345: Weird behavior with links at the end of the line in blockquotes
If you add a link to the end of a line in a blockquote, the link is pushed to a new line. Creating a new line after the link creates a new, separate blockquote.
### Steps to reproduce
1. Start a new post in the visual editor.
2. Start a blockquote with regular text.
3. Add a link at the end of the line of text. Result: The link is moved to a new line.
4. Hit enter to start a new line. Result: The new line starts in a new blockquote.

I tried using the link button to create the link and link text, and I tried typing the link text and selecting it to create the link, with the same results.

Screencast: https://cloudup.com/c7nA73baZ1E

aforcier:
Fixes #345. The cause was weird WebView behavior when calling `execCommand('insertHTML')` to insert the link tag: instead of being inserted within the current `div`, it was added outside of it.

So instead of:

``` HTML
<blockquote>
  <div>text <a href="http://example">link</a></div>
</blockquote>
```

we had:

``` HTML
<blockquote>
  <div>text </div>
  <a href="http://example">link</a>
</blockquote>
```

This also caused autocorrect issues, as usually happens when adding text anywhere but inside a `div`.

The fix was to wrap links in `div` tags inside blockquotes, but I had to exclude a few cases: empty newline inside blockquote, and inserting a link in the middle of a line (with text surrounding on both sides). Those were cases where the `div` wrapping is unnecessary, and causes other issues if used (double-wrapping in `div`s, or toggling `div` off on the entire line).

cc @maxme

### DIFF
--- a/libs/editor-common/assets/ZSSRichTextEditor.js
+++ b/libs/editor-common/assets/ZSSRichTextEditor.js
@@ -818,12 +818,16 @@ ZSSEditor.insertHTMLWrappedInParagraphTags = function(html) {
     this.insertHTML(paragraphOpenTag + space + paragraphCloseTag);
 };
 
-// Needs addClass method
-
 ZSSEditor.insertLink = function(url, title) {
     var html = '<a href="' + url + '">' + title + "</a>";
 
-    if (this.getFocusedField().getHTML().length == 0) {
+    var parentBlockQuoteNode = ZSSEditor.closerParentNodeWithName('blockquote');
+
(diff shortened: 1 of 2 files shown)


maxme:
Confirmed fix for #345. :shipit: 

Note: I found a unrelated minor issue while testing and opened https://github.com/wordpress-mobile/WordPress-Editor-Android/issues/368

PR #364, opened 2016-04-28
Issue #227: Buggy paragraph formatting when exiting lists

### LINKED ISSUE #227: Blockquote wraps list above text
When turning off a list and the caret is after a word already in a `<li>`, blockquoteing that word will cause the list to be blockquoted as well. This video demonstrates it better than I can describe:

https://cloudup.com/c2aW80mp-0L

If the `<li>` item is blank, the blockquote works as expected.

ref: #226

aforcier:
Fixes #227. The ultimate cause of the bug was that exiting a list by toggling off the list button didn't wrap the resulting text in paragraph tags. This caused blockquote wrapping to include the list above it, and also led to autocorrect issues.

Specifically, starting with (as shown in the video in the issue):

``` HTML
<div>
  <ol>
    <li>Item 1</li>
    <li>Item 2</li>
  </ol>
</div>
```

Pressing the list button with the cursor on Item 2 resulted in:

``` HTML
<div>
  <ol>
    <li>Item 1</li>
  </ol>
</div>
Item 2
```

With this fix, we now get:

``` HTML
<div>
  <ol>
    <li>Item 1</li>
  </ol>
</div>
<div>
  Item 2
</div>
```

And blockquote wrapping after that works correctly (only wraps 'Item 2').

cc @maxme

### DIFF
--- a/libs/editor-common/assets/ZSSRichTextEditor.js
+++ b/libs/editor-common/assets/ZSSRichTextEditor.js
@@ -2606,8 +2606,24 @@ ZSSEditor.completeListEditing = function() {
             if (node.nodeType == 1 &&
                     (node.tagName.toUpperCase() == NodeName.UL
                         || node.tagName.toUpperCase() == NodeName.OL)) {
-                // Make a new P node as a sibling to the parent node
-                document.execCommand('insertParagraph', false);
+
+                var focusedNode = document.getSelection().getRangeAt(0).startContainer;
+
+                if (focusedNode.nodeType == 3) {
+                    // If the focused node is a text node, the list item was not empty when toggled off
+                    // Wrap the text in a div and attach it as a sibling to the div wrapping the list
+                    var parentParagraph = focusedNode.parentNode;
+                    var paragraph = document.createElement('div');
+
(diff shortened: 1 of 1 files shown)


libs/editor-common/assets/ZSSRichTextEditor.js
```diff
@@ -2606,8 +2606,24 @@ ZSSEditor.completeListEditing = function() {
             if (node.nodeType == 1 &&
                     (node.tagName.toUpperCase() == NodeName.UL
                         || node.tagName.toUpperCase() == NodeName.OL)) {
-                // Make a new P node as a sibling to the parent node
-                document.execCommand('insertParagraph', false);
+
+                var focusedNode = document.getSelection().getRangeAt(0).startContainer;
+
+                if (focusedNode.nodeType == 3) {
```
maxme, line 2612: Note: if we have to clean ZSSEditor someday, it would be great to use const static var for node types, even if we know what the 3 is for, it's always better for reading the code ;).

maxme:
Confirmed fix for #227. :shipit:

PR #366, opened 2016-04-29
Issue #365: IndexSizeError when inserting links in empty posts

### LINKED ISSUE #365: IndexSizeError in JS log when inserting a link in an empty post
Log:

```
E/WordPress-EDITOR: Uncaught IndexSizeError: Failed to execute 'setEnd' on 'Range': There is no child at offset 1. -- From line 432 of file:///android_asset/ZSSRichTextEditor.js
```

aforcier:
Fixes #365. The error was because `ZSSEditor.canExpandBackward`/`canExpandForward` is being passed a `div` node, rather than the text node it expects, when the post is empty. Returning `false` automatically avoids the index out of bounds error.

cc @maxme

### DIFF
--- a/libs/editor-common/assets/ZSSRichTextEditor.js
+++ b/libs/editor-common/assets/ZSSRichTextEditor.js
@@ -411,29 +411,31 @@ ZSSEditor.getSelectedText = function() {
 };
 
 ZSSEditor.canExpandBackward = function(range) {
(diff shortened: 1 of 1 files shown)


maxme:
Looks good :shipit: (and apparently, I have to fix my whitespace editor config for JS files)

PR #372, opened 2016-05-06
Issue #301: Tap to delete media

### LINKED ISSUE #301: Media: Add support for tapping to remove media from post
In alpha testing, Derek noted that it would be helpful to have an 'x' or something similar to tap to remove media from a post (instead of backspacing):

> Adding images/video was easy, but I had trouble figuring out how to remove them. It would be great if there was an ‘x’ or something that I could tap to remove.

For comparison, the visual web editor has a popup that appears when you click on an image in the editor, with an 'x' to remove the image:

screen shot 2016-02-25 at 1 04 15 pm

aforcier:
Fixes #301, adding a 'delete' button to the overlay when an image is selected:

301-tap-to-delete-media

Tapping the delete button removes the image from the post.

<h4>Notes</h4>

- Does not work for video. Videos don't have a settings screen, and adding an overlay when tapping them is tricky since tapping plays/pauses the video. A solution I see for this is to permanently overlay a delete button on the top right of videos. I'll be opening a separate issue.
- I moved all the inline SVGs in the editor CSS (edit buttons) to standalone SVG files for better readability

cc @maxme and @mattmiklic for design feedback


libs/editor-common/assets/editor.css
```diff
@@ -143,12 +143,12 @@ span.img_container.failed::before {
     -webkit-user-select: none;
     pointer-events: none;
     content:"";
-    background:url('data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg" width="54" height="54" viewBox="0 0 124 124"><circle fill="none" stroke="#FFFFFF" stroke-width="4"  cx="62" cy="62" r="60"/><path d="M79.0637899,40.93125 C74.1838649,36.0375 67.4465291,33 59.9831144,33 C45.0562852,33 33,45.0825 33,60 C33,74.9175 45.0562852,87 59.9831144,87 C72.5628518,87 83.0994371,78.39375 86.0881801,66.75 L79.0637899,66.75 C76.2776735,74.61375 68.8142589,80.25 59.9831144,80.25 C48.7879925,80.25 39.7204503,71.188125 39.7204503,60 C39.7204503,48.811875 48.7879925,39.75 59.9831144,39.75 C65.5722326,39.75 70.587242,42.07875 74.2514071,45.740625 L63.3602251,56.625 L87,56.625 L87,33 L79.0637899,40.93125 L79.0637899,40.93125 Z" fill="#FFFFFF"/></svg>') no-repeat center top;
+    background:url('svg/retry-image.svg') no-repeat center top;
```
maxme: +1

mattmiklic:
Tried a couple of ideas to help the "remove" icon better match the existing Edit icon, so they feel more related. I don't think red is necessary here since it's not destructive (no data is being deleted), the user can always re-add the photo to the post. 

Mockup:
mockup

I added a 10% black background to the icons, which I think would make sure they're always visible, even on light images. 

Measurements:
screen shot 2016-05-06 at 5 56 42 pm

I've exported the `gridicons-cross` and `gridicons-pencil` icons at the needed sizes here: https://cloudup.com/ceuV4iy2USZ

aforcier:
That looks much better :grinning:.

One request: could I get the icons in `svg` format since that's how things are set up within the visual editor (they're loaded in the CSS)?

Also what are your thoughts on updating the retry button along with the others?

187-media-upload-fail-overlay

All the current assets (as of this branch) are here, for reference:

https://github.com/wordpress-mobile/WordPress-Editor-Android/tree/7b6c732360e8d5ab98cf07e133aaba92f2aaa2c2/libs/editor-common/assets/svg

There are two sizes for each of the retry and edit icons in there, for different sizes of images.

mattmiklic:
> One request: could I get the icons in svg format since that's how things are set up within the visual editor (they're loaded in the CSS)?

Absolutely. Actually, if you need SVGs we can just use the originals direct from Gridicons:

`gridicons-cross`:
https://github.com/Automattic/gridicons/blob/master/svg-min/gridicons-cross.svg

`gridicons-pencil`:
https://github.com/Automattic/gridicons/blob/master/svg-min/gridicons-pencil.svg

mattmiklic:
Just in case you need the SVGs already colored and resized I've re-exported them here, too. https://cloudup.com/cwFmnARhgOi

aforcier:
301-edit-overlay-white

<h3>Done:</h3>
- [x] Changed the edit pencil icon from dashicon to gridicon
- [x] Switched to adding the circle border around edit icon in CSS rather than inside the SVG itself
- [x] Changed close button style to match @mattmiklic's specs
- [x] Added `10%` dark background to edit and delete icons
- [x] Adjusted the font of the `Edit` text to Medium (was set to weight 600, it's now 500)
- [x] Adjusted edit button for small images (less than `100dp` on any side), it's now the same as the normal sized edit button (but without a caption or the circle border)

<h3>Remaining:</h3>
- [x] Switch retry button from dashicon to gridicon
- [x] Add `10%` dark background to retry button
- [x] Update retry icon for large images accordingly
- [x] Fix backspace to delete images that have their edit overlay visible 

Re: the last point, I found that backspacing an image that's selected (so that the edit overlay is visible) leaves the container HTML in the document. This seems to be caused by adding the `delete-overlay` element to the `edit-container`, and the fix will probably require tracking the edit overlay with the `MutationObserver`.

cc @mattmiklic for another look

mattmiklic:
Looking good to me now. 👍

aforcier:
Updated:

372-retry-gridicon
- [x] Retry button for failed media (normal and large) is now a gridicon, and I applied the same style changes as for the edit button
- [x] Fixed bug with manually deleting selected images

aforcier:
Also fixed a bug where selecting an image > pressing edit > returning from the edit settings screen > tapping the same image would show the edit settings screen again, instead of showing the edit overlay.

maxme:
New remove button works well, but I think the Edit image settings screen doesn't save anymore, I'll check without the patch.

maxme:
This commit 69f55745a0c1522e7dfb266a1f9c43bc6453abc3 is the culprit, apparently it's too early to set `currentEditingImage` to null here, update image will be called later and image edits won't be saved.

aforcier:
> This commit 69f5574 is the culprit, apparently it's too early to set currentEditingImage to null here, update image will be called later and image edits won't be saved.

Fixed in https://github.com/wordpress-mobile/WordPress-Editor-Android/pull/372/commits/b250069b1d1be53e35629550e1a0eab7546d7d2e. You were right, it was too early, [`updateCurrentImageMeta()`](https://github.com/wordpress-mobile/WordPress-Editor-Android/blob/b250069b1d1be53e35629550e1a0eab7546d7d2e/libs/editor-common/assets/ZSSRichTextEditor.js#L1900) relies on `currentEditingImage` when the settings dialog is closed and changes are saved. `updateCurrentImageMeta()` clears `currentEditingImage` when it's done; the problem was that when the dialog is dismissed without saving the ZSSEditor isn't notified. That should work correctly now.

maxme:
:shipit:

PR #378, opened 2016-05-19
fix #376: escape quotes in imageMeta var

### LINKED ISSUE #376: Strings containing quotes (or double quotes) in the image settings editor aren't saved
We probably need to escape them when calling `ZSSEditor.updateCurrentImageMeta`.

https://cloudup.com/icAGIKvKams

maxme:
fix #376: escape quotes in imageMeta var


aforcier:
:shipit:

PR #379, opened 2016-05-19
fix #368: remove a workaround that introduced the issue

### LINKED ISSUE #368: Links can be tapped several times in a row resulting in multiple "edit link" pop over opened
See https://cloudup.com/c0KBjL2Daua

maxme:
Problem comes from the 500ms delay [here](https://github.com/wordpress-mobile/WordPress-Editor-Android/blob/develop/libs/editor-common/assets/ZSSRichTextEditor.js#L3264-L3266). I'm not sure to understand why there is a workaround here. Does `after-tap` mean `tap up` event?

Looking at `git blame`, I guess that was used in iOS. When I remove the delay, everything seems to work as expected.

### DIFF
--- a/libs/editor-common/assets/ZSSRichTextEditor.js
+++ b/libs/editor-common/assets/ZSSRichTextEditor.js
@@ -3256,14 +3256,8 @@ ZSSField.prototype.handleTapEvent = function(e) {
         if (targetNode.nodeName.toLowerCase() == 'a') {
             var arguments = ['url=' + encodeURIComponent(targetNode.href),
                              'title=' + encodeURIComponent(targetNode.innerHTML)];
-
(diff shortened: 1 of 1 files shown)


aforcier:
Looks like the `500ms delay was added in https://github.com/wordpress-mobile/WordPress-Editor-iOS/pull/283, to fix a bug with an edge case with updating links: https://github.com/wordpress-mobile/WordPress-Editor-iOS/issues/220. I think the idea was to force the link-tap callback to occur after the tap event bubbling had completed, because that was somehow causing issues in the iOS WebView.

I tried to reproduce that iOS bug on Android running this branch, but I wasn't able to on API 16, 18, or 23. Given that, and that it fixes the actual issue with multiple edit dialogs, :shipit:.

PR #381, opened 2016-05-20
Fix #380: Add isAdded() checks where getActivity is called

### LINKED ISSUE #380: Monkey crash report in editor 1.0
Looks like a missing `isAdded()` check (we should check other `getActivity()` calls)

```
// java.lang.NullPointerException
//  at org.wordpress.android.editor.EditorFragment$5.run(EditorFragment.java:458)
//  at java.lang.Thread.run(Thread.java:856)
```

maxme:
Fix #380: Add isAdded() checks where getActivity is called


aforcier:
:shipit:

PR #384, opened 2016-05-20
Issue #373: Don't scroll to the current cursor position when tapping an image

### LINKED ISSUE #373: Tap an image to show the "Edit" button, edit view will scroll to cursor
1. Add N images to a post (without text between them)
2. Tap the first image
3. Notice the view scrolls to the cursor (end of post in that case)

**or** watch this: https://cloudup.com/c7dqAS2PGsZ

aforcier:
Fixes #373.

When an image was tapped with the keyboard down, the screen would scroll to the current cursor position as the keyboard came up, which could be very far from the image that was actually tapped.

With this patch, the cursor is moved to the image, so the screen only scrolls to the bottom of the image that was tapped when the keyboard comes up.

cc @maxme

### DIFF
--- a/libs/editor-common/assets/ZSSRichTextEditor.js
+++ b/libs/editor-common/assets/ZSSRichTextEditor.js
@@ -417,6 +417,19 @@ ZSSEditor.giveFocusToElement = function(element, offset) {
     selection.addRange(range);
 };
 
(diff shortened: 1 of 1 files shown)


maxme:
:shipit:

PR #386, opened 2016-05-25
fix #382: show a more graphic in place of the more tag

### LINKED ISSUE #382: Show more tags in the visual editor
Currently, more tags aren't displayed at allin the visual editor. Eventually we want to re-implement the more button for the visual editor, perhaps as part of an ['advanced' format bar](https://github.com/wordpress-mobile/WordPress-Editor-Android/issues/371), but in the meantime we should add some sort of visual affordance for the more tag. This is what the Calypso visual editor uses:

screen_shot_2016-05-05_at_7 44 20_am

maxme:
fix #382: show a more graphic in place of the more tag


maxme:
Update with the more graphic:
screenshot-2016-05-27_16 07 20 476

davewhitley:
Hm, what would be the best way to go about it? Would you do some 9 scale magic on it, to fill the entire width on large screens? Is it possible to build this in code without an svg? I'm thinking of how it would be done it CSS — a dotted border element. It would be nice to have "More" actual text for translation and such.

maxme:
> Is it possible to build this in code without an svg?

Probably doable with multiple html elements+ some css, but the svg solution seems easier.

> It would be nice to have "More" actual text for translation and such.

I'm not sold on this, since the only method to add it currently is to use `<!--more-->` tag. Also great to have the same behavior in Calypso Web and mobile.

rachelmcr:
> > It would be nice to have "More" actual text for translation and such.
> 
> I'm not sold on this, since the only method to add it currently is to use  tag. Also great to have the same behavior in Calypso Web and mobile.

Agreed. It might be nice to someday have the word "More" in the visual editor translated, but Calypso also uses the English word "More" even in other languages. It's nice to match the behavior between Calypso web and mobile, and perhaps someday we can improve that everywhere.

aforcier:
Found an issue when using a hardware keyboard (reproducible on an API23 device with a real hardware keyboard, and an API16 emulator with hardware keyboard). When backspacing, the line under the more tag becomes wrapped in the more `span`, rather than the more tag being deleted. Afterwards, pressing enter duplicates the more tag:

https://cloudup.com/cAqN96aFVci

This is the HTML when backspace is used and the 'Strikethrough' text is wrapped in the more `span`:

``` HTML
<div><span class="more-tag" wp-more-data=" c'est mort ">Strikethrough<br></span></div>
```

This also happens with non-formatted text (it's not caused by the strikethrough tags).

Everything works normally on both devices when using a software keyboard.

Edit: This also happens with the nextpage tag, unsurprisingly.

libs/editor-common/assets/example-content.html
```diff
@@ -7,7 +7,7 @@
 
 <a href="http://www.wordpress.com">I'm a link!</a>
 
-<!--more-->
+<!--more c'est mort -->
```
aforcier, line 10: 👍

davewhitley:
I disagree about the Calypso consistency. If we have an opportunity to improve something we shouldn't be held back. It should be fixed on the web side too, but no one has gotten around to it. I'm sure that the read more tag in Calypso was just hastily put in there without much thought in an effort to ship sooner.

davewhitley:
Here's svg http://cl.ly/2L2a1O3e100a

It's colored black, but should be colored using one of the color variables. Let me know if that works.

maxme:
> I'm sure that the read more tag in Calypso was just hastily put in there without much thought in an effort to ship sooner.

We're in the same situation, we ship the editor next monday, we'll use the current more/nextpage graphics and revisit later.

maxme:
> . When backspacing, the line under the more tag becomes wrapped in the more span

@aforcier This is fixed in 53cd44c

aforcier:
> > I'm sure that the read more tag in Calypso was just hastily put in there without much thought in an effort to ship sooner.
> 
> We're in the same situation, we ship the editor next monday, we'll use the current more/nextpage graphics and revisit later.

Agreed, we'll open a separate issue to address this in the future.

aforcier:
With the 53cd44c fix everything's working well! :shipit:

aforcier:
Opened a separate issue for graphics/translation improvements: #388.

Issue #398, opened 2016-06-04
Buggy blockquotes at start of post on older APIs (API 16-19)


aforcier:
When starting a post with a blockquote, the cursor isn't indented properly. It seems that the cursor isn't correctly placed inside the opening div, which also causes a second div to be added when the user begins typing (this leads to the blockquote splitting in two when pressing enter and other strangeness).

PR #399, opened 2016-06-04
Issue #385: Autocorrect blockquote bug

### LINKED ISSUE #385: Buggy blockquote behavior at start of post
When starting a post with a blockquote, typing in text and pressing enter will remove the blockquote under the following conditions:
- API 23+
- Default system keyboard with autocorrect enabled
- Only one word has been typed and no whitespace characters have been added yet
- The word before the cursor is going to be modified by autocorrect (e.g. it's misspelled)

The underlying reason why this is happening is unclear. I initially thought this was an issue introduced by a recent Android System WebView update, but it seems to actually be a system keyboard issue (see notes). The deepest source of the problem I could find relates to strange behavior with the Enter `keyCode` under the above circumstances.

In [`handleKeyDownEvent`](https://github.com/wordpress-mobile/WordPress-Editor-Android/blob/8e571a6a276ebe0a2c961f6444c42e4c0a0a82dd/libs/editor-common/assets/ZSSRichTextEditor.js#L3144), the event `keyCode` is captured and some special handling is done. Most normal keys entered with the software keyboard appear as a [`229` keyCode](https://lists.w3.org/Archives/Public/www-dom/2010JulSep/att-0182/keyCode-spec.html), which makes it impossible for the software keyboard to identify the key typed until it's already been entered. One of the exceptions (normally) is the enter key, which has `keyCode` `13`.

For some reason, on API23 (and Android N Preview) the enter key registers as `229` instead of `13` when pressed inside a `blockquote` or `li` element, and this is keeping some of our special handling for enter key from working [`here`](https://github.com/wordpress-mobile/WordPress-Editor-Android/blob/8e571a6a276ebe0a2c961f6444c42e4c0a0a82dd/libs/editor-common/assets/ZSSRichTextEditor.js#L3188). Note that this only happens if autocorrect is enabled and is targeting a word (i.e. there is some normal text immediately before the cursor).

<h4>Notes</h4>
- I initially thought this was a WebView version issue, but it's not - the bug occurs regardless of System WebView version on API23, and never occurs (also regardless of System WebView version) on API22. I think it's probably a system keyboard issue.
- Not reproducible using hardware keyboard or using the SwiftKey keyboard
- Not reproducible with auto-correct turned off in the keyboard settings
- Not reproducible on emulators - the system keyboard behaves differently on emulators, sending the correct `keyCode`s for all keys pressed

aforcier:
Fixes #385.

This addresses two separate issues:
1. When inside a blockquote that's not at the start of a post, pressing enter and triggering autocorrect will add an extra break tag: https://github.com/wordpress-mobile/WordPress-Editor-Android/issues/385#issuecomment-222795524
2. When inside a blockquote at the start of a post, pressing enter and triggering autocorrect will remove the blockquote: https://github.com/wordpress-mobile/WordPress-Editor-Android/issues/385#issue-156114496

These only seem to occur on API23+, using the default system keyboard with autocorrect enabled. If the currently highlighted word won't be changed by autocorrect when pressing enter, everything works normally.

cc @maxme

### DIFF
--- a/libs/editor-common/assets/ZSSRichTextEditor.js
+++ b/libs/editor-common/assets/ZSSRichTextEditor.js
@@ -3431,13 +3431,25 @@ ZSSField.prototype.afterKeyDownEvent = function(beforeHTML, e) {
         // Give focus to new div
         var newFocusElement = focusedNode.firstChild;
         ZSSEditor.giveFocusToElement(newFocusElement, 1);
-    } else if (focusedNode.nodeName == NodeName.DIV && focusedNode.parentNode.nodeName == NodeName.BLOCKQUOTE
-        && focusedNode.parentNode.previousSibling == null && focusedNode.parentNode.childNodes.length == 1
-        && focusedNodeIsEmpty) {
-        // When a post begins with a blockquote, and there's content after that blockquote, backspacing inside that
-        // blockquote will work until the blockquote is empty. After that, backspace will have no effect
-        // This fix identifies that situation and makes the call to setBlockquote() to toggle off the blockquote
-        ZSSEditor.setBlockquote();
+    } else if (focusedNode.nodeName == NodeName.DIV && focusedNode.parentNode.nodeName == NodeName.BLOCKQUOTE) {
+        if (focusedNode.parentNode.previousSibling == null && focusedNode.parentNode.childNodes.length == 1
+            && focusedNodeIsEmpty) {
+            // When a post begins with a blockquote, and there's content after that blockquote, backspacing inside that
+            // blockquote will work until the blockquote is empty. After that, backspace will have no effect
+            // This fix identifies that situation and makes the call to setBlockquote() to toggle off the blockquote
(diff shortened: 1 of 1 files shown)


aforcier:
Since I found [this independent blockquote issue](https://github.com/wordpress-mobile/WordPress-Editor-Android/issues/398) on older APIs and it obscures whether this patch has any side effects of its own, I'm adding the 'Do Not Merge' label until I fix #398 and can fully test the effects of this patch on those API levels.

aforcier:
I've opened a PR (https://github.com/wordpress-mobile/WordPress-Editor-Android/pull/400) which addresses #398. Once that PR is tested and merged I'll pull the changes into this branch and do a full re-test - hopefully the two combined will correct all 'blockquote at start of post' issues we have (as well as item 1 above).

aforcier:
With #400 merged, I've pulled `develop` in and did a re-test, this is ready for review now @maxme.

maxme:
:shipit:

PR #402, opened 2016-06-07
Issue #401: Blockquote toggle off issues in new posts

### LINKED ISSUE #401: Issues with toggling empty blockquote off at start of post
When inserting a blockquote in an empty post and immediately pressing the blockquote button again, with no text entered, a few strange things happen.
1. On API 16-19, the blockquote is not removed. On an API19 emulator I was able to remove the blockquote by using backspace, but on my API18 Samsung S3 I can't remove the blockquote that way either (I have to enter text, then toggle off the blockquote, then deleted the text).
2. On API 21+, the blockquote is removed but leaves behind an empty space above the cursor. Ideally, the placeholder text would reappear instead.

aforcier:
Fixes #401, correcting the behavior when toggling off an empty blockquote when the post is otherwise empty.

To test, launch editor with empty post and toggle blockquote on and then off again without typing anything.
1. On API19 and below, toggling the blockquote off now actually works, removing the empty blockquote
2. On all API levels, toggling the blockquote off no longer leaves an empty newline behind, so that now the placeholder shows up again right away

cc @maxme

### DIFF
--- a/libs/editor-common/assets/ZSSRichTextEditor.js
+++ b/libs/editor-common/assets/ZSSRichTextEditor.js
@@ -688,6 +688,7 @@ ZSSEditor.setBlockquote = function() {
     var selection = document.getSelection();
     var range = selection.getRangeAt(0).cloneRange();
     var sendStyles = false;
+    var doEmptyFieldCleanup = false;
 
     // Make sure text being wrapped in blockquotes is inside paragraph tags
     // (should be <blockquote><paragraph>contents</paragraph></blockquote>)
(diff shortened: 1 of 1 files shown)


maxme:
> On all API levels, toggling the blockquote off no longer leaves an empty newline behind, so that now the placeholder shows up again right away

I still get the empty new line on API 23. (and API 21)

aforcier:
> I still get the empty new line on API 23. (and API 21)

Fixed in 5664496, the cause was overzealous refactoring 😅.

maxme:
Looks good! :shipit:

PR #406, opened 2016-06-08
Escape invisible unicode separator characters in HTML mode

### LINKED ISSUE #405: Handle U+2028 invisible line separator character in HTML editor
From https://github.com/wordpress-mobile/WordPress-Editor-iOS/pull/827, the Android editor also doesn't handle `U+2028` (invisible line separator) when switching to HTML and back.

aforcier:
Fixes #405.

To test, copy and paste each of the following lines into the editor (there are invisible separators, `U+2028` between the two 'Line' words, and `U+2029` between the two 'Paragraph' words):

```
Line Line

Paragraph Paragraph
```

Next, switch to HTML mode and back. Without this patch, any changes made during HTML mode won't be preserved, and there should be a console error when switching back to visual from HTML mode:

```
E/WordPress-EDITOR: Uncaught SyntaxError: Unexpected token ILLEGAL -- From line 1 of 
```

With this patch, there should be no error and changes made in HTML mode should be preserved.

cc @maxme

### DIFF
--- a/WordPressEditor/src/main/java/org/wordpress/android/editor/Utils.java
+++ b/WordPressEditor/src/main/java/org/wordpress/android/editor/Utils.java
@@ -61,6 +61,11 @@ public class Utils {
             html = html.replace("'", "\\'");
             html = html.replace("\r", "\\r");
             html = html.replace("\n", "\\n");
+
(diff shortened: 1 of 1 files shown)


maxme:
:shipit:

PR #407, opened 2016-06-09
Sync WPAndroid

aforcier:
`subtree push` of WPAndroid's `develop`, with updated example app version to fix Travis.

cc @maxme


maxme:
:shipit:

Issue #409, opened 2016-06-11
Difficulty setting the cursor under large media


aforcier:
When uploading media, we automatically add a new paragraph underneath. However, switching to HTML mode or closing and reopening the post clears away the new paragraph, and this can make it difficult to tap the screen and set the cursor to the right of or below the media item.

We can probably just add a rule to `ZSSEditor.applyVisualFormatting()` to add a new paragraph when a post ends in a media item.

Issue #408, opened 2016-06-11
Videopress videos continously attempt to load


aforcier:
When uploading a video to a .COM site, closing and re-opening the post or switching to HTML mode and back to visual before the video has finished processing will cause the editor to continuously attempt to load the video and poster image (before they're available), causing some glitchy UI behavior as well:

https://cloudup.com/c1rrUT1R3Nb

Issue #296, opened 2016-02-25
Unable to delete a video using backspace on some devices


aforcier:
**Update**: With recent video fixes, the issue now only applies to the HTC One (M8) (`5.1.0`), and only to backspacing videos at the start of posts.

<h5>Original issue:</h5>

Deleting a video using the backspace key on a virtual keyboard doesn't work on XPeria Z5 (`5.1.1`) or the HTC One (M8) (`5.1.0`).

Unable to reproduce on emulators running `4.2.2`, `5.0`, `5.1.0`.

Original ref: https://github.com/wordpress-mobile/WordPress-Android/pull/3782#issuecomment-188350641

maxme:
Tested on my XPeria Z5 (now `6.0`), and I was able to delete the video. Maybe it was fixed by the `p2div` branch?

aforcier:
@tonyr59h: could you give this another try on your HTC One? If it's not reproducible there either we can probably close this.

tonyr59h:
Local video uploads-in-progress delete as expected with the virtual keyboard backspace.

Videos that have finished uploading and videos added from the WordPress media library delete as expected **_unless**_ the video element is the first element in the post.

For example, backspacing the video works for:

```
This is a post
[video code]
...
```

but does not work for:

```
[video code]
...
```

Behaves as expected on the 5X minus some cursor positioning hiccups.

maxme:
I can repro what Tony's describing on the z5 Android 6.0: When a post starts with `[wpvideo code]` it can't be deleted in visual mode with the delete key.

maxme:
Debug HTML, with this test case:

```
<div contenteditable="true" id="zss_field_content" class="field" placeholdertext="Partagez ici votre histoire..." style="min-height: 383px;"><div><br></div><div><video webkit-playsinline="" src="https://taliwutblog.files.wordpress.com/2016/04/wp-1461675661880.mp4" onclick="" controls="controls" preload="metadata" poster="https://taliwutblog.files.wordpress.com/2016/04/wp-1461675661880_std.original.jpg" data-wpvideopress="hmbzwjhI"></video></div></div>
```

aforcier:
[...]

PR #410, opened 2016-06-12
Issue #377: Video remove button

### LINKED ISSUE #377: Video: Add support for tapping a "X" button to remove a video from post
Like what was done for images in https://github.com/wordpress-mobile/WordPress-Editor-Android/pull/372

aforcier:
Fixes #377 and #408.

Adds a permanent span container to videos, with a delete button:

377-video-delete-overlay

This PR also fixes an issue reported by @tonyr59h, that adding remote videos to an empty post makes it difficult to tap under the video afterward. The problem was that the editor wasn't adding a newline after those videos, like it does for all images and for uploading videos.

The best way to test is to `subtree pull` this branch into WPAndroid, but it's also possible to check the remote video insertion code in isolation by calling [`ZSSEditor.insertVideo()`](https://github.com/wordpress-mobile/WordPress-Editor-Android/blob/531f3687115620564764e0c1a94f5875438cb151/libs/editor-common/assets/ZSSRichTextEditor.js#L1538) in the Chrome dev console with appropriate arguments.

There's a more general issue with paragraph formatting around media, which applies to all media and is due to the newline under media being stripped when switching to HTML mode and back, or opening an existing post. That's being tracked in #409.

Hopefully this also fixes all issues with #296 (needs testing on the Z5 and the HTC One now that `span`-wrapped videos are in the picture).

cc @maxme, and also @tonyr59h for a runthrough since he's the most talented at summoning video bugs

### DIFF
--- a/libs/editor-common/assets/ZSSRichTextEditor.js
+++ b/libs/editor-common/assets/ZSSRichTextEditor.js
@@ -1658,6 +1658,12 @@ ZSSEditor.replaceLocalVideoWithRemoteVideo = function(videoNodeIdentifier, remot
         containerNode.replaceWith(videoNode);
     }
 
+    // Wrap video in edit-container node for a permanent delete button overlay
+    var containerHtml = '<span class="edit-container"><span class="delete-overlay" contenteditable="false"></span></span>';
+    videoNode.insertAdjacentHTML('beforebegin', containerHtml);
+    var selectionNode = videoNode.previousSibling;
+    selectionNode.appendChild(videoNode);
+
     var joinedArguments = ZSSEditor.getJoinedFocusedFieldIdAndCaretArguments();
     ZSSEditor.callback("callback-input", joinedArguments);
     // We invoke the sendVideoReplacedCallback with a delay to avoid for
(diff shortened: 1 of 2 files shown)


maxme:
The button looks good and works well but apparently it breaks the playback (pulled the subtree in wpandroid): https://cloudup.com/c6RwYZBACc8

maxme:
It works better on the example app https://cloudup.com/cAf7fHN8VyA but in both tests, I had the cursor moved before the X button after tapping the video.

tonyr59h:
> I had the cursor moved before the X button after tapping the video.

Saw the same in the example app.

aforcier:
> The button looks good and works well but apparently it breaks the playback (pulled the subtree in wpandroid): https://cloudup.com/c6RwYZBACc8

This turns out to be a pre-existing issue: https://github.com/wordpress-mobile/WordPress-Editor-Android/issues/408.

> It works better on the example app https://cloudup.com/cAf7fHN8VyA but in both tests, I had the cursor moved before the X button after tapping the video.

This should be fixed in 0acaa82 - I extended `contenteditable="false"` to apply to the whole container, not just the delete overlay. With that change I can't reproduce the issue anymore.

Ready for another look @maxme 😄

aforcier:
Updated with a fix for  #408 (flickering videopress videos when switching to HTML mode and back right after uploading a videopress video).

maxme:
Looks good, one issue noticed https://github.com/wordpress-mobile/WordPress-Editor-Android/issues/415 but it's not related to this changeset.

:shipit:

Issue #393, opened 2016-06-01
Crash report 5.4: Cannot call method 'insertBefore' of null


maxme:
```
Non-fatal Exception: org.wordpress.android.JavaScriptException: Uncaught TypeError: Cannot call method 'insertBefore' of null
       at JavaScriptException.(ZSSRichTextEditor.js:1971)
```

maxme:
resolved by https://github.com/wordpress-mobile/WordPress-Android/issues/4157

PR #414, opened 2016-06-15
Issue #411: Edit containers not being cleared when switching to HTML mode

### LINKED ISSUE #411: Edit overlay can get "stuck" on an image
### Expected behavior:

No matter what steps I take, if I tap an image to bring up the "Edit" overlay, I expect to be able to disable that overlay and not have any consequences on the published post.
### Actual behavior:

If I tap an image to bring up the "Edit" overlay, and then I switch to HTML view, exit the editor, or publish the post, the "Edit" overlay gets stuck over the image and publishes the word "Edit" in the post.

Tapping the image again brings up a second "Edit" overlay on top of the first one.
### Steps to reproduce:
1. Open the visual editor to start a new post
2. Add an image
3. Tap the image to bring up the Edit overlay
4. Switch to the HTML view
5. Switch back to the visual view. Result: The "Edit" overlay can't be dismissed
6. Tap the image again. Result: A second "Edit" overlay appears on the image
7. Publish the post. Result: The word "Edit" is added to the post, above the image (from `<span class="edit-content">Edit</span>`)
#### Screenshots

After switching to HTML view and back, and tapping on the text in the post (with the intention of dismissing the Edit overlay):

Double Edit overlay:

aforcier:
Fixes #411.

This bug was introduced fairly recently, in https://github.com/wordpress-mobile/WordPress-Android/pull/4157/commits/82903c73310d3af5e05cf74f2b583f10fde19cb6. That commit was fixing a crash report (https://github.com/wordpress-mobile/WordPress-Editor-Android/issues/393), but it turns out that the null check will trip when switching to HTML mode and the edit container isn't cleared away. I changed the null check behavior so that #393 should still be fixed.

cc @maxme

### DIFF
--- a/libs/editor-common/assets/ZSSRichTextEditor.js
+++ b/libs/editor-common/assets/ZSSRichTextEditor.js
@@ -1984,11 +1984,6 @@ ZSSEditor.applyImageSelectionFormatting = function( imageNode ) {
 }
 
 ZSSEditor.removeImageSelectionFormatting = function( imageNode ) {
-    if (!$('#zss_field_content')[0].contains(imageNode)) {
-        // The image node has already been removed from the document
-        return;
-    }
-
     var node = ZSSEditor.findImageCaptionNode( imageNode );
     if ( !node.parentNode || node.parentNode.className.indexOf( "edit-container" ) == -1 ) {
         return;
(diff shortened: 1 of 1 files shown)


maxme:
:shipit:

PR #420, opened 2016-06-20
Issue #409: Difficulty setting the cursor under media

### LINKED ISSUE #409: Difficulty setting the cursor under large media

aforcier:
Fixes #409. When loading a post (either by loading from DB or by switching from HTML mode), the editor will now append a new paragraph after media items, allowing the user to easily place the cursor under the media item.

This should have no effect on posts being published in WPAndroid, as any trailing spaces or break tags are stripped by the editor when extracting the post from the visual editor.

cc @maxme


maxme:
:shipit:

PR #422, opened 2016-06-20
Issue #421: Failed videos aren't cleared correctly when switching to HTML mode

### LINKED ISSUE #421: Failed videos aren't cleared when switching to HTML mode
When switching to HTML mode with failed videos in the post, the 'Remove failed uploads?' prompt is shown. If 'Remove failed uploads' is selected, the editor switches to HTML mode, but the failed video upload remains in the post, container and all.

aforcier:
Fixes #421. The problem was that `ZSSEditor.removeAllFailedMediaUploads()` was only attempting to remove images, so any failed video HTML would be left in the post.

cc @maxme


maxme:
:shipit:

PR #424, opened 2016-06-22
Issue #423: Can't backspace to delete images with captions

### LINKED ISSUE #423: Cannot backspace to delete image with caption
<ol>
    <li>New post</li>
    <li>Insert an image</li>
    <li>Add caption to image</li>
    <li>Place cursor immediately after image</li>
    <li>Press backspace on keyboard </li>
</ol>

Keyboard is dismissed, the cursor is stuck halfway between the caption and the image. If you bring up the keyboard again and try backspace again, weird stuff happens (last word of the caption is added as regular text to the post but inline with the caption).

aforcier:
Fixes #423.

For some reason, the cursor position gets reset when backspacing into an image with a caption. I could not reproduce the issue on emulators (using hardware or software keyboard), and I think this is another autocorrect+WebView bug.

The best fix I found was to make captions editable (previously they had `contenteditable="false"`). This allows backspace to focus on the caption, and backspacing after the caption is deleted will remove the image as expected (I guess this is actually a feature 🎉).

I also extended the changes of https://github.com/wordpress-mobile/WordPress-Editor-Android/pull/418 and https://github.com/wordpress-mobile/WordPress-Editor-Android/pull/420 to captioned images in this PR (properly wrapping media in paragraphs when switching from HTML to visual, and adding an empty space under media under the same circumstances).

The `wp.loadText` method we're using doesn't treat captions the same way as other media, so a few extensions had to be added (in `editor-utils-formatter.js`) to treat captions the same way (3f15fa3). I also extended the JS tests to check for this.

cc @maxme

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,9 @@ With npm and Mocha installed, from within `libs/editor-common/assets/test`, run:
 
     npm install chai
 
-It should now be possible to run the tests using `mocha` inside `libs/editor-common/assets`.
+And then run `mocha` inside `libs/editor-common/assets`:
+
+    cd libs/editor-common/assets; mocha; cd -
 
 ## LICENSE ##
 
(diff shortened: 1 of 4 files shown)


aforcier:
Updated to remove space that was being added in front of the caption text (predates this PR).

maxme:
Tested in wpandroid on API 16, 21, 23.
:shipit: 

Note: I added c04f955 so I could quickly copy/paste the test command (adding it to the gradle build would be better, but this project will be put in maintenance mode soon).

PR #425, opened 2016-06-23
Fix broken image link in example post

aforcier:
Fixes the broken image link in the example post.

cc @maxme


maxme:
Cowboy cat is back 🐮 😼  :shipit:

Issue #417, opened 2016-06-16
Apply formatting to currently selected word


aforcier:
When pressing the inline format buttons (bold, italic, strikethrough), the behavior should be to apply the style to the currently selected word. This would mirror the behavior on Calypso, and also hopefully sidestep the inline styling autocorrect issue: https://github.com/wordpress-mobile/WordPress-Editor-Android/issues/123.

Also, starting a new line should clear inline styles, again mirroring Calypso.

aforcier:
Added in #427.

PR #427, opened 2016-06-24
Issue #417: Apply inline formatting to current word

aforcier:
Addresses #417. Using in-line formatting (bold, italics, strikethrough) with the cursor inside a word will apply the formatting to the entire word.

417-expand-inline-formatting

This PR also fixes an issue in the original selection expansion code (written for adding links to text): https://github.com/wordpress-mobile/WordPress-Editor-Android/commit/5031c47b2dec8c479176237bd709767c0f773eca. When some text was already selected, pressing the link button would still expand to the entire word, making it impossible to apply a link to part of a word without using HTML mode.

This also essentially fixes the mid-word styling autocorrect bug #123 (by avoiding it entirely). Autocorrect doesn't interfere with applying the styling to the whole word at once, and the only way to apply formatting to part of a word now is to select that part first and then press the format button, which is also not interfered with by autocorrect.

<h3>A few issues/notes</h3>

- Using a format button will leave the word highlighted at the end (demonstrated in the GIF above). I made some attempts to reset the cursor position but I wasn't able to get consistent results due to the way the formatting action changes the DOM. I'm not sure how big of an issue this is, but I suggest addressing this as a separate issue if we decide to change it.
- Using a format button (say, bold) at the start or end of a word will apply formatting to the entire word. This differs from what a user might expect to happen (and happens on Calypso, for example), which is that bold will be turned on but will only affect text about to be typed. Unfortunately we can't support that functionality due to the autocorrect bug (#123). My take on this is that if a user expected to format the entire word they'll be pleased (but only if we keep this PR as is), but if they expected it to turn on and stay on when they press space they'll be disappointed anyway.

cc @maxme


maxme:
> Using a format button will leave the word highlighted at the end (demonstrated in the GIF above). I made some attempts to reset the cursor position but I wasn't able to get consistent results due to the way the formatting action changes the DOM. I'm not sure how big of an issue this is, but I suggest addressing this as a separate issue if we decide to change it.

I like it, make it more obvious the whole word was selected/changed.

aforcier:
> > Using a format button will leave the word highlighted at the end (demonstrated in the GIF above). I made some attempts to reset the cursor position but I wasn't able to get consistent results due to the way the formatting action changes the DOM. I'm not sure how big of an issue this is, but I suggest addressing this as a separate issue if we decide to change it.
> 
> I like it, make it more obvious the whole word was selected/changed.

👍 I was thinking that too

maxme:
:shipit: 

> This PR also fixes an issue in the original selection expansion code (written for adding links to text): 5031c47. When some text was already selected, pressing the link button would still expand to the entire word, making it impossible to apply a link to part of a word without using HTML mode.

Fix confirmed.

PR #428, opened 2016-06-27
Issue #204: Can't start a post with formatted text

### LINKED ISSUE #204: Can't start a post with formatted text
Pressing **B** or **I** in a new post and then typing will turn the style button back off and won't apply it.

Doesn't look to be an `API 19+` autocorrect issue, could reproduce on an `API18` Samsung S3.

Demo video: https://cloudup.com/cPrhIt_F0DW

Original ref: https://github.com/wordpress-mobile/WordPress-Android/pull/3018#issuecomment-125737183

aforcier:
Fixes #204.

The issue was that our start-of-post paragraph manipulations were clearing away any formatting that the user had enabled before starting to type. A manual storing + re-applying the formatting did the trick.

cc @maxme


maxme:
:shipit:

PR #435, opened 2016-07-05
Issue #432: evaluateJavaScript not supported

### LINKED ISSUE #432: Crash report 5.5: API not supported on Android 4.3 and earlier
```
Fatal Exception: java.lang.IllegalStateException: This API not supported on Android 4.3 and earlier
       at android.webkit.WebViewClassic.evaluateJavaScript(WebViewClassic.java:2669)
       at android.webkit.WebView.evaluateJavascript(WebView.java:894)
       at org.wordpress.android.editor.EditorWebView.execJavaScriptFromString(EditorWebView.java:17)
       at org.wordpress.android.editor.EditorFragment$1$1.run(EditorFragment.java:169)
       at android.os.Handler.handleCallback(Handler.java:733)
       at android.os.Handler.dispatchMessage(Handler.java:95)
       at android.os.Looper.loop(Looper.java:136)
       at android.app.ActivityThread.main(ActivityThread.java:5017)
       at java.lang.reflect.Method.invokeNative(Method.java)
       at java.lang.reflect.Method.invoke(Method.java:515)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:779)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:595)
       at dalvik.system.NativeStart.main(NativeStart.java)
```

aforcier:
Fixes #432. The issue occurs in custom ROMs using the 4.3 'classic' WebView (pre-Chromium) in a Android 4.4 build. Fixed by reverting to the 'compatibility' WebView in those cases, which uses the older `loadUrl` method for executing JavaScript (as well as applying our reflection-based modifications to avoid the keyboard being dismissed on JS execution).

cc @maxme

### DIFF
--- a/WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragment.java
+++ b/WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragment.java
@@ -26,6 +26,7 @@ import android.view.ViewGroup;
 import android.view.inputmethod.InputMethodManager;
 import android.webkit.URLUtil;
 import android.webkit.WebView;
+import android.widget.RelativeLayout.LayoutParams;
 import android.widget.ToggleButton;
 
 import com.android.volley.toolbox.ImageLoader;
@@ -154,6 +155,7 @@ public class EditorFragment extends EditorFragmentAbstract implements View.OnCli
             int index = parent.indexOfChild(mWebView);
             parent.removeView(mWebView);
             mWebView = new EditorWebViewCompatibility(getActivity(), null);
+            mWebView.setLayoutParams(new LayoutParams(LayoutParams.MATCH_PARENT, LayoutParams.MATCH_PARENT));
             parent.addView(mWebView, index);
         }
 
--- a/WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragment.java
+++ b/WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragment.java
@@ -148,6 +148,15 @@ public class EditorFragment extends EditorFragmentAbstract implements View.OnCli
 
         mWebView = (EditorWebViewAbstract) view.findViewById(R.id.webview);
 
+        // Revert to compatibility WebView for custom ROMs using a 4.3 WebView in Android 4.4
+        if (mWebView.shouldSwitchToCompatibilityMode()) {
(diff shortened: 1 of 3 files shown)


aforcier:
Note: The crash can be simulated by using `EditorWebView` instead of `EditorWebViewCompatibility` [here](https://github.com/wordpress-mobile/WordPress-Editor-Android/blob/260a0d5ba0eae70b44d4ccc826d2a2025f344280/WordPressEditor/src/main/res/layout/editor_webview.xml).

However, doing this on a `4.3` emulator will result in a `NoSuchMethodError` rather than an `IllegalStateException`. The custom ROMs are using a pre-release KitKat snapshot (as far as I can tell), which actually [implemented `evaluteJavascript`](https://android.googlesource.com/platform/frameworks/base/+/1eb8b726d311df7e41e6bdf839a1401ef0a9c741/core/java/android/webkit/WebViewClassic.java#2662) but it's a stub that throws the `IllegalStateException` we're seeing in the crash log.

I opted to catch both errors just in case.

WordPressEditor/src/main/java/org/wordpress/android/editor/EditorFragment.java
```diff
@@ -148,6 +148,15 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
 
         mWebView = (EditorWebViewAbstract) view.findViewById(R.id.webview);
 
+        // Revert to compatibility WebView for custom ROMs using a 4.3 WebView in Android 4.4
+        if (mWebView.shouldSwitchToCompatibilityMode()) {
+            ViewGroup parent = (ViewGroup) mWebView.getParent();
+            int index = parent.indexOfChild(mWebView);
+            parent.removeView(mWebView);
+            mWebView = new EditorWebViewCompatibility(getActivity(), null);
```
maxme, line 157: I think we should add `mWebView.setLayoutParams(new LayoutParams(LayoutParams.MATCH_PARENT, LayoutParams.MATCH_PARENT));
` here

aforcier, line 157: Good catch, added in be1722d.

maxme:
:shipit:

Issue #383, opened 2016-05-20
Drop the WordPress-Editor-Common sub-project


aforcier:
It ~~might be~~ is time to abandon the idea of a shared core library with the iOS editor and do away with the complexity of maintaining a [symlink](https://github.com/wordpress-mobile/WordPress-Editor-Android/blob/develop/WordPressEditor/src/main/assets) in the Android editor project.

Discussion here: https://github.com/wordpress-mobile/WordPress-Android/pull/4083#issuecomment-220331563

aforcier:
Done in #436.

PR #436, opened 2016-07-06
Issue #383: Drop WordPress-Editor-Common

aforcier:
Addresses #383, getting rid of `libs/editor-common` in the editor, and moving the assets to the main project folder.

I also did some cleanup, moving all the libraries we shouldn't be modifying to a `libs/` folder inside `assets/`.

Note: we'll also need to update the WPAndroid README to delete the symlink instructions for Windows when this is merged.

cc @maxme


maxme:
Lib dependency paths must be updated in `test-formatter.js` (eg. `require('../libs/underscore-min.js');` instead of `require('../underscore-min.js');` )

maxme:
It's not a new thing, but now that I think of it, we should not have these `example` and `test` directories (and `node_modules` subdir) in the main `assets/` directory (everything in there will be packaged in the final apk).

`example` should be moved to the example project, and `test` could be moved in `WordPressEditor/src/test/js`

aforcier:
> Lib dependency paths must be updated in test-formatter.js (eg. require('../libs/underscore-min.js'); instead of require('../underscore-min.js'); )

Oops, I was sure I had run the tests before pushing :( Fixed

> example should be moved to the example project, and test could be moved in WordPressEditor/src/test/js

Agreed, done in 7072c80 and c017777. I updated the README with new instructions (note that now you need to run `mocha test*` rather than just `mocha`, that allowed me to drop the `test/` subfolder (`test/js/test/` was getting too awkward).

Ready for another pass @maxme!

maxme:
:shipit:

PR #450, opened 2016-08-26
Fix #445: translated alignement issue in ImageSettings screen

### LINKED ISSUE #445: Image alignment class in image source is translated in non-English languages
### Expected behavior

When I set an image's alignment in the image settings, I expect it to add the class `alignleft`, `aligncenter`, or `alignright` on the image to control how the image is aligned in the post.
### Actual behavior

If the device is set to a non-English language, the alignment value is translated and so the image isn't correctly aligned in the post. For example, if the device/interface language is set to Spanish, the image ends up with the class `alignizquierda`, `aligncentrado`, or `alignderecha`.
### Steps to reproduce
1. In the device settings, select a non-English language
2. In the app, start a new post
3. Add an image to the post
4. Open the image settings and select a left/center/right alignment
5. Save the image settings
6. Switch to HTML view and view the image source

maxme:
Fix #445: translated alignement issue in ImageSettings screen

Note: once this is merged in develop, this should be subtree pulled in wpandroid `release/5.8` and the strings from the editor copied to `strings.xml` in wpandroid.

### DIFF
--- a/WordPressEditor/src/main/java/org/wordpress/android/editor/ImageSettingsDialogFragment.java
+++ b/WordPressEditor/src/main/java/org/wordpress/android/editor/ImageSettingsDialogFragment.java
@@ -32,7 +32,6 @@ import org.wordpress.android.util.MediaUtils;
 import org.wordpress.android.util.ToastUtils;
 
 import java.util.Arrays;
-import java.util.Locale;
 import java.util.Map;
 
 /**
@@ -52,6 +51,7 @@ public class ImageSettingsDialogFragment extends DialogFragment {
     private EditText mCaptionText;
     private EditText mAltText;
     private Spinner mAlignmentSpinner;
+    private String[] mAlignmentKeyArray;
     private EditText mLinkTo;
     private EditText mWidthText;
     private CheckBox mFeaturedCheckBox;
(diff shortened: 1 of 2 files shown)


WordPressEditor/src/main/java/org/wordpress/android/editor/ImageSettingsDialogFragment.java
```diff
@@ -159,12 +159,9 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
...
 
                 String alignment = mImageMeta.getString("align");
-
-                // Capitalize the alignment value to match the spinner entries
-                alignment = alignment.substring(0, 1).toUpperCase(Locale.US) + alignment.substring(1);
-
-                String[] alignmentArray = getResources().getStringArray(R.array.alignment_array);
-                mAlignmentSpinner.setSelection(Arrays.asList(alignmentArray).indexOf(alignment));
+                mAlignmentKeyArray = getResources().getStringArray(R.array.alignment_key_array);
+                int alignementIndex = Arrays.asList(mAlignmentKeyArray).indexOf(alignment);
```
aforcier: Minor typo here, `s/alignementIndex/alignmentIndex`

maxme: Fixed in    88577e2

aforcier:
:shipit:

